PR #3, opened 2018-10-26
Fix parser issues found testing the official Lua testsuite.

anyska:


kyren:
Added testsuite in c78bdf87c18fe8dedd11836628f8a42938e26495

I just parse every Lua file from the official lua testsuite and nothing else.  Obviously these should eventually be run and tested in some better way, but probably when I get to that point the official testsuite examples won't be as useful (or at least they won't be useful for quite a long time probably).

anyska:
This is what I was using to test `luster`:
```rust
fn main() {
    for entry in fs::read_dir(env::args().nth(1).unwrap()).unwrap() {
        let entry = entry.unwrap();
        let path = entry.path();
        if path.extension() == Some(OsStr::new("lua")) {
            let src = fs::read(&path).unwrap();
            if let Err(chunk) = parse_chunk(&src[..]) {
                println!("{:?} : {}", path.file_stem().unwrap(), chunk);
            }
        }
    }
}
```
As for `gll`, we're not doing any performance benchmarking yet, but instead checking whether `luster` and `gll-lua` produce matching parse trees (see [WIP code](https://github.com/lykenware/gll/blob/lua/lua/src/check_luster.rs)).
So far, everything we tested matches `luster`, so that's a success! 
(We still have to get all testcases parsing with `gll-lua`, including non-UTF8 and the really large ones)

PR #5, opened 2018-10-31
Fix possible string slice length panic

Veykril:
While actually running the test suites just to see what happens I noticed a panic that occurs due to the `copy_from_slice` function asserting that both slices are of the same length.


kyren:
Ugh, I left this for a bit and then totally forgot about it.  You are absolutely correct, thanks for the fix!

PR #7, opened 2019-02-04
First attempt at a REPL using Luster.

OmnipotentEntity:


examples/repl.rs
```diff
@@ -0,0 +1,79 @@
...
+                    println!("{}", out_string);
+                    collecting = false;
+                }
+                Ok(None) => {
+                    prompt = ">> ";
+                    line.push_str("\n"); // separate input lines
+                }
+                Err(err_string) => {
+                    // let line_clone = line.clone();
+                    // editor.add_history_entry(line_clone);
```
kyren: You forgot some dead commented code

examples/repl.rs
```diff
@@ -0,0 +1,88 @@
...
+                                    .map(|value| format!("{:?}", value))
+                                    .collect::<Vec<_>>()
+                                    .join("\t");
+                                Ok(Ok(output))
+                            }
+                            Err(Error::ParserError(ParserError::EndOfStream { expected: e })) => {
+                                // continue reading input and append it to `line`
+                                Ok(Err(Error::ParserError(ParserError::EndOfStream {
+                                    expected: e,
+                                })))
```
kyren: you can use @ patterns to make this a bit easier

OmnipotentEntity: Thanks for the tip.

kyren:
One thing this makes me realize is that it would be nice if sequences didn't *have* to return `Result`.  I blatantly copied the design of futures 0.2, what I *should* have done was copy the design of futures 0.3 where that's not true.  I'll do this soon hopefully when I refactor the `Sequence` trait again.

kyren:
This is pretty good, and I think it's important to have a repl for people to play with to get a sense of how far along or not far along the interpreter is.

I know that I told you (out of band) that it should be an example to reduce dependencies, but now I'm thinking 1) no, and 2) it should be joined with the main interpreter so that when people run `cargo run` they get a repl.

You don't have to do that part, or you can if you'd like but it doesn't have to be in the same PR, whichever way you want.  I can merge as is if you'd like

kyren:
This is great, I'm gonna merge this and probably move the compiler / parser to the examples folder, as they're really just debugging tools, and that way it's the only binary and you can run it with `cargo run`.  We could also possibly add them as modes to the main binary like you suggested earlier.  But this is great, thanks!

PR #8, opened 2019-02-07
Binary Operators

OmnipotentEntity:


src/value.rs
```diff
@@ -113,6 +120,99 @@ impl<'gc> Value<'gc> {
...
+        Some(Value::safe_div(a, b, &|a, b| {
+            Value::Number((a / b).floor())
+        }))
+    }
+
+    // When given a % b, lua computes the remainder, not the modulo.
+    // However, Rust computes the modulo correctly.  (e.g. -2 % 3 = 1 according to lua, and -2
+    // according to Rust.)
+    // This is why there is the second step.  Hopefully, the compiler will optimize the extra
+    // mod out
```
kyren: I wonder if this behavior is dependent on the C compiler / architecture used to compile Lua, I bet it is.

src/value.rs
```diff
@@ -113,6 +120,99 @@ impl<'gc> Value<'gc> {
         }
     }
 
+    // A small helper function to handle division-like zero handling
```
kyren: Small nit, this doesn't need to be a method on Value, this and `copysign` can go to the bottom of the file since they're internal helper functions.

src/value.rs
```diff
@@ -113,6 +120,99 @@ impl<'gc> Value<'gc> {
...
+            // Seems that all nans are negative in lua
+            (ref a, ref b) if a.is_zero() && b.is_zero() => Value::Number(-f64::NAN),
+            (ref a, ref b) if b.is_zero() => {
+                Value::Number(copysign(f64::INFINITY, a.to_f64().unwrap()))
+            }
+            (a, b) => f(a, b),
+        }
+    }
+
+    // This operation always returns a Number, even when called by int arguments
```
kyren: The documentation on all the pub functions should be doc comments (though none of the doc comments in the whole project right now are really very *good* documentation, they're intended as at least placeholder doc comments.

src/value.rs
OmnipotentEntity: I'm hoping you're wrong, but I know you're right.

src/compiler/operators.rs
```diff
@@ -126,6 +126,62 @@ pub fn simple_binop_opcode(
...
+            (RegisterOrConstant::Register(left), RegisterOrConstant::Constant(right)) => {
+                OpCode::PowRC { dest, left, right }
+            }
+            (RegisterOrConstant::Constant(left), RegisterOrConstant::Register(right)) => {
+                OpCode::PowCR { dest, left, right }
+            }
+            (RegisterOrConstant::Constant(left), RegisterOrConstant::Constant(right)) => {
+                OpCode::PowCC { dest, left, right }
+            }
+        },
```
kyren: Quick sanity check, is the total number of opcodes after you add all of the binary and unary operators > 255?  I don't think it is, or is even close, but if it is it might be an issue (it will cause opcodes to take > 4 bytes).

src/value.rs
OmnipotentEntity: I will look up proper doc formatting and update this then.

src/opcode.rs
```diff
@@ -286,4 +286,92 @@ pub enum OpCode {
...
+        right: ConstantIndex8,
+    },
+    UnegR {
+        dest: RegisterIndex,
+        value: RegisterIndex,
+    },
+    UnegC {
+        dest: RegisterIndex,
+        value: ConstantIndex8,
+    },
```
kyren: Quick sanity check, is the total number of opcodes after we add all of the binary and unary operators > 255? I don't think it is, or is even close, but if it is it might be an issue (it will cause opcodes to take > 4 bytes).

src/value.rs
```diff
@@ -113,6 +120,99 @@ impl<'gc> Value<'gc> {
...
+    pub fn float_divide(self, other: Value<'gc>) -> Option<Value<'gc>> {
+        let (a, b) = match (self, other) {
+            (Value::Integer(a), Value::Integer(b)) => (a as f64, b as f64),
+            (Value::Number(a), Value::Number(b)) => (a, b),
+            (Value::Integer(a), Value::Number(b)) => (a as f64, b),
+            (Value::Number(a), Value::Integer(b)) => (a, b as f64),
```
kyren: It seems like there should be some `as_number` Value cast which returns an `Option<f64>`... or something better than this, there seem to be a lot of these as f64 castas and checking whether both sides can be expressed as numbers.

src/opcode.rs
OmnipotentEntity: Currently there are 70 opcodes.

src/value.rs
OmnipotentEntity: Yes, I was thinking that, it would also help with String coercion down the line.  I just wasn't sure if now was the correct time to start that.

src/value.rs
kyren: It doesn't have to be PROPER, just add three '///' is fine for now

src/value.rs
kyren: Fun fact, Lua number parsing depends on the system locale.  This is why the target compatibility at the language level is "aggressive" not "complete".

src/value.rs
```diff
@@ -113,6 +120,99 @@ impl<'gc> Value<'gc> {
...
+        f: &Fn(T, T) -> Value<'gc>,
+    ) -> Value<'gc> {
+        match (lhs, rhs) {
+            // Seems that all nans are negative in lua
+            (ref a, ref b) if a.is_zero() && b.is_zero() => Value::Number(-f64::NAN),
+            (ref a, ref b) if b.is_zero() => {
+                Value::Number(copysign(f64::INFINITY, a.to_f64().unwrap()))
+            }
+            (a, b) => f(a, b),
+        }
```
kyren: This is prooobably not best expressed as a match, `if a.is_zero() && b.is_zero() { ... } else if b.is_zero() { ... } else { ... }` is better.

OmnipotentEntity: That's fair.

src/value.rs
kyren: No, it doesn't depend on the C compiler, it's guaranteed to have C99 behavior: https://github.com/kyren/rlua/blob/e351bf57fa4c3985109d6674a8b6c35f9890da16/lua/lvm.c#L572

I don't *like* C99 behavior, but it is consistent.

Cargo.toml
```diff
@@ -15,3 +15,4 @@ num-traits = "0.2"
 gc-arena = { path = "./gc-arena" }
 rustyline = "3.0.0"
 clap = "~2.32"
+num = "*"
```
kyren: Why the addition of depending on num?  I don't think it's used anywhere?

src/value.rs
OmnipotentEntity: I'm glad to hear it's defined.  I'm noticing that you're attempting to avoid overflow in rlua, but the PUC Rio engine doesn't seem to even allow you to construct an Integer with values INT_MIN.

> -9223372036854775808
-9.2233720368548e+18
> -9223372036854775807
-9223372036854775807

Cargo.toml
OmnipotentEntity: Unfortunately, it's used for its traits so that I didn't have to write an integer and a float version of safe_div.  Specifically, I'm using "Zero" and "ToPrimitive".

OmnipotentEntity: oooh, I think I see what you're getting at, yeah, I should just pin to num-traits, and not pull in all of num.

OmnipotentEntity: aaaand, num-traits is already in the dependencies.  Well, I feel somewhat foolish now.

src/value.rs
kyren: That file I linked you isn't rlua code, rlua embeds Lua 5.3.5, so I just linked you to the PUC-Rio Lua source inside it.

Or are you referring to rlua's behavior with `Value::Integer` not guarding against INT_MIN?

OmnipotentEntity: Oh, of course.  You can't directly create a MIN_INT value, but one can arise as part of a calculation, so that check of course still makes sense.

\> -9223372036854775807 - 1  
-9223372036854775808  
\> (-9223372036854775807 - 1) * -1  
-9223372036854775808  
\> (-9223372036854775807 - 1) % 1  
0

kyren: .....wat

kyren: Well, okay now that I look at it it's not quite wat material, Lua allows you to observe all the weird integer overflow behavior as long as the literals you type all are integers, so that *kind of* makes sense... I haven't dug this deep into Lua's numeric behavior before.

src/opcode.rs
kyren: Okay so should be okay since there are not >= 47 more binary operators to implement (4 opcodes each)

src/value.rs
uazu: > I wonder if this behavior is dependent on the C compiler / architecture used to compile Lua, I bet it is.

No, see the Lua docs. Lua uses floor division, and modulus is defined similarly. This is better mathematically, as most repeating things don't have a discontinuity at 0!

uazu: Okay, I see you found this. Sorry to pop up in your commit BTW. I must have 'watched' this!

OmnipotentEntity:
Current known bug:

> 2/3
Integer(0)
> 2//3
Integer(2)

I'm looking into it.

src/value.rs
OmnipotentEntity: This as been addressed by the functions `to_number`, `to_integer` and `bin_op`.

src/value.rs
```diff
@@ -79,50 +80,120 @@ impl<'gc> Value<'gc> {
...
+            |a, b| safe_div(a, b, |a, b| Value::Integer(a.wrapping_div(b))),
+            |a, b| safe_div(a, b, |a, b| Value::Number((a / b).floor())),
+        )
+    }
+
+    /// When given a % b, lua computes the remainder, not the modulo.
+    /// However, Rust computes the modulo correctly.  (e.g. -2 % 3 = 1 according to lua, and -2
+    /// according to Rust.)
+    /// This is why there is the second step.  Hopefully, the compiler will optimize the extra
+    /// mod out
```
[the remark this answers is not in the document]
OmnipotentEntity: Of course I got those two things backwards.  Because I am bad at math.

src/vm.rs
```diff
@@ -457,7 +457,7 @@ pub fn step_vm<'gc>(
                 } => {
                     let left = registers.stack_frame[left.0 as usize];
                     let right = registers.stack_frame[right.0 as usize];
-                    if (left < right) == skip_if {
+                    if (left < right) != skip_if {
```
OmnipotentEntity: So I fixed the issue with the less than operator.  But I don't like how I fixed it.  Because I don't understand how it can be possible that this is different from the equality operator.  You said out of thread that even though `== skip_if`is different from Lua, it ought to be fine as long as its consistent.  Which makes me think that there's some bug in the compiler generating something backwards, and I undid the backward bug, rather than fixing the issue the right way.  But I've spent a few hours looking at it, and fiddling with it, but I can't find it.

kyren: The compiler is probably reversing the arguments, you wouldn't see a bug with equals because it's symmetric

src/compiler/operators.rs
```diff
@@ -213,6 +273,36 @@ pub fn comparison_binop_opcode(
                 }
             }
         },
+        ComparisonBinOp::LessThan => match (left, right) {
+            (RegisterOrConstant::Register(left), RegisterOrConstant::Register(right)) => {
+                OpCode::LessRR {
+                    skip_if: !skip_if,
```
kyren: you're inverting skip_iff here when you don't want to

src/vm.rs
kyren: Never mind, found it, see above review comment

src/compiler/operators.rs
OmnipotentEntity: Thanks!

src/value.rs
```diff
@@ -79,50 +80,124 @@ impl<'gc> Value<'gc> {
...
+        // No need for special casing, 0^0 = 1 in both Rust and Lua
+        bin_op(
+            self,
+            other,
+            |a, b| Value::Number((a as f64).powf(b as f64)),
+            |a, b| Value::Number(a.powf(b)),
+        )
+    }
+
+    pub fn unary_negate(self) -> Option<Value<'gc>> {
```
kyren: I think just 'negate' is fine

kyren:
I know that you're not done with *all* of the binary operators, but this patch is pretty good right now, the only thing it's missing is renaming `unary_negate` to just `negate`, and using `read_integer`, `read_hex_integer`, `read_float`, and `read_hex_float` from the lexer module to support decimal / hexadecimal string -> numeric conversions.  Don't worry about moving those methods from lexer or doing any re-org like I said before, I think they're already public in the lexer module.

I think the simplest way is just to try parsing as hex, then decimal, since parsing as hex already has the 0x / 0X check.

I'd like to merge after that if that's okay?  It's easier to review smaller PRs

src/value.rs
```diff
@@ -79,50 +83,126 @@ impl<'gc> Value<'gc> {
...
+        if self.to_number().is_some() && other.to_number() == Some(0.0) {
+            return Some(Value::Number(-f64::NAN));
         }
+
+        bin_op(
+            self,
+            other,
+            |a, b| safe_div(a, b, |a, b| Value::Integer(((a % b) + b) % b)),
+            |a, b| safe_div(a, b, |a, b| Value::Number(((a % b) + b) % b)),
+        )
```
kyren: As per discussion, `safe_div` doesn't really buy much here, and this will potentially parse a number twice.

tests/running/math.lua
```diff
@@ -0,0 +1,125 @@
...
+           0.0 /  0   ~=  0 / 0 and
+           1   // 0   ~= -1 / 0 and
+           1.0 // 0.0 ~= -1 / 0 and
+           1   // 0.0 ~= -1 / 0 and
+           1.0 // 0   ~= -1 / 0 and
+           0   // 0   ~=  0 / 0 and
+           0.0 // 0.0 ~=  0 / 0 and
+           0   // 0.0 ~=  0 / 0 and
+           0.0 // 0   ~=  0 / 0
+end
```
kyren: As per discussion, an `is_nan` helper function might make this a little clearer

kyren:
I'm sorry for the chaos I am causing to your PR branch

PR #9, opened 2019-02-13
Bitwise operators and comparison operators

OmnipotentEntity:


kyren:
This will pretty much finish the compiler / bytecode MVP! :tada:

Neither is perfect, there are still some missing improvements to both even to bring it into full parity with PUC-Rio Lua 5.3, but after this all of it exists in some form.

I think the only things left for perfect parity are supporting more than 64k constants in a function (Lua supports 256k), constructing tables in a more efficient way with something like OP_SETLIST, and table pre-sizing in OP_NEWTABLE.  There might be something else I'm forgetting...

PR #10, opened 2019-02-15
Math std lib

OmnipotentEntity:
Pushing incomplete math library.  Untested so far.  Only implemented functions starting with A-L.  Watch this space.  Need to figure out arguments and so on.


kyren:
I'm going to go ahead and merge this before changing the callback API to allow for immediate results again.

PR #11, opened 2019-02-19
Added other math functions (except random)

OmnipotentEntity:
and cleaned up argument handling (now abs(2, 3) is an error).  This also includes some minor bug fixes such as floor and ceil returning a Number instead of an Integer.

I'm somewhat concerned about how max and min are handled.  It feels very complicated for what is essentially a fold.  I might be missing some way of properly handling this sort of thing.  I couldn't make the ? operator work like I was expecting.

Tests to come soon (TM)


kyren:
I don't think there are any Lua stdlib functions that check for extra arguments.

```
> math.acos(0.0, "foo")
1.5707963267949
```

kyren:
Regarding max / min, you can either not use a fold at all, or you can use `Iterator::try_fold`.

kyren:
Also behavior of min / max with no arguments:

```
> math.min()
stdin:1: bad argument #1 to 'min' (value expected)
```

src/stdlib/base.rs
```diff
@@ -79,4 +79,29 @@ pub fn load_base<'gc>(mc: MutationContext<'gc, '_>, root: LuaRoot<'gc>, env: Tab
...
+                    b"Missing argument to type",
+                )))
+                .into());
+            }
+            match args.get(0).cloned().unwrap() {
+                // Lua5.3 does not differentiate between a floating point and integer number
+                Value::Integer(_) => 
+                    Ok(CallbackResult::Return(vec![Value::String(
+                                    String::new_static(b"number"),
+                                    )])),
```
OmnipotentEntity: I'm somewhat worried about this function in particular.  Is this a decent way to implement it considering that lua just throws away the integer vs number distinction?  Or should I try something else?

kyren:
Just curious, why `rand_xoshiro` specifically?

src/stdlib/base.rs
kyren: I don't super like that you have to specify the string "number" twice, I wonder if type_name itself should be changed?

kyren: Yeah, probably just change type_name to match, I think that's acceptable and simpler.

kyren:
Okay, something I'm going to be working on soon is changing both the argument and return types to callbacks to avoid creating intermediate buffers.  There will be a type that callbacks can't create that serves as both the argument and return vector, with operations to move args / returns around in the vector, and callbacks will be required to return it.

This is going to be an involved refactor or every callback, would you like me to make this change before or after merging this PR?

src/stdlib/base.rs
```diff
@@ -79,4 +79,47 @@ pub fn load_base<'gc>(mc: MutationContext<'gc, '_>, root: LuaRoot<'gc>, env: Tab
...
+    .unwrap();
+
+    env.set(
+        mc,
+        String::new_static(b"select"),
+        Callback::new_immediate(mc, |args| {
+            match args.get(0).cloned().unwrap_or(Value::Nil).to_integer() {
+                Some(n) if n >= 1 && (n as usize) <= args.len() => Ok(CallbackResult::Return(
+                    args[n as usize..args.len()].to_vec(),
+                )),
```
kyren: I don't think this is right, isn't this off by one?  n == 1 will miss the first argument?

OmnipotentEntity: Nope, because the arguments to select are `select(number_to_select, other_args...)` so n == 1 will give the arguments from index 1 (the first of `other_args`) to the end.

OmnipotentEntity:
`rand_xoshiro` was chosen because it is very fast and has decent rng properties.  It is not unique by any stretch of the imagination, it's just the built in random function I thought was best when I glanced once briefly over the built-in engines in the [rand crate documentation](https://rust-random.github.io/book/guide-rngs.html).

OmnipotentEntity:
> This is going to be an involved refactor or every callback, would you like me to make this change before or after merging this PR?

Either way is fine, I'm OK with changing the callbacks to a new convention myself, or you can at your option.  The only thing that's keeping this PR from being complete is tests for functions alphabetically beyond `frexp`.  And if I'm keeping you from working on something that needs to get done, just do it, and I'll catch up.  This code isn't the most complicated in the world. :)

src/stdlib/base.rs
kyren: oh right, duh

OmnipotentEntity:
Hopefully, this ought to be the rest of the needed tests.  There are a few edge cases that I'm aware of where behavior of the functions does not line up 1 to 1 with lua, such as the min function does not handle nan cases in quite the same way.  And some functions will accept strings with numbers in them where lua just barfs on it (such as the max and min functions), I'm not sure if you want to adhere to the behavior of PUC Lua on this because the documentation and specification do not say that this should be either allowed or disallowed.

kyren:
I think it's important that all tests pass both luster and PUC-Rio Lua, so if there are any tests that aren't like that we shouldn't codify the current mismatched behavior of luster.

It's fine if it's not perfectly compatible right now in the interests of simple implementations, because there are bigger compatibility fish to fry first, but if you want to leave commented out tests that don't match, that would be excellent.

OmnipotentEntity:
All test cases work in both luster and PUC-Rio Lua.  I'll add some tests and comment them out where the behavior is mismatched.

tests/running/math.lua
```diff
@@ -215,6 +215,8 @@ function test17()
            math.max(3, 3.0, 3.0) == 3 and
            is_integer(math.max(3, 3.0, 3.0)) and
            math.max(-5, -4, -3, -2, -1, 0, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1) == 10
+    -- Tests we do not currently pass due to incompatibility with PUC-Rio Lua
+    --     is_err(math.max("1"))
```
kyren: In PUC-Rio Lua 5.3.5, it seems like `math.max("1")` returns `1`, does luster return error here?

tests/running/math.lua
```diff
@@ -225,6 +227,9 @@ function test18()
            math.min(3, 3.0, 3.0) == 3 and
            is_integer(math.min(3, 3.0, 3.0)) and
            math.min(5, 4, 3, 2, 1, 0, -10, -9, -8, -7, -6, -5, -4, -3, -2, -1) == -10
+    -- Tests we do not currently pass due to incompatibility with PUC-Rio Lua
+    --     is_err(math.min("1"))
+    --     is_nan(math.min(0.0 % 0.0, 1, 2))
```
kyren: I'm pretty sure `math.min` with a nan argument is pretty much undefined, as `math.min(nan, 1.0)` and `math.min(1.0, nan)` are different in PUC-Rio Lua.

OmnipotentEntity: They are different.  The code matches behavior with `math.min(1.0, nan)` but not with `math.min(nan, 1.0)`

tests/running/math.lua
OmnipotentEntity: Sorry, I didn't actually check the behavior of that, I just assumed it errored.  In actuality, it's `math.max(1, "2", 1)` that errors in PUC-Rio Lua.  I'll adjust it.

kyren:
I lied about refactoring arguments / returns, I can't come up with a sane API atm and I'm stalled working on it.

I'm going to go ahead and merge this PR.

PR #14, opened 2019-03-04
add `luster --dump` flag to print the parsed bytecode

poga:
First of all, thanks for the amazing project! I learned a lot from it. :)

This PR added a new command line flag `--dump`(`-d`). When provided, luster will print the parsed bytecode without actually running the code. I find it pretty useful for understanding the internal. It might also be helpful for others.

For now, I just dump the `FunctionProto` with `Debug` trait with minimal formatting. It's definitely note a good textual representation of luster's bytecode. However, it's fine for learning purpose IMO.

Example:

```lua
-- test.lua
local x = 1

function foo(y)
    return x + y
end

print(foo(2))
```

```
$ luster --dump test.lua
=============
FunctionProto(0x7ffee596b9d0)
=============
fixed_params: 0, has_varargs: true, stack_size: 4
constants:
0: Integer(1)
1: String(Short8(3, Gc { ptr: [102, 111, 111, 0, 0, 0, 0, 0] }))
2: String(Short8(5, Gc { ptr: [112, 114, 105, 110, 116, 0, 0, 0] }))
3: Integer(2)
opcodes:
0: LoadConstant { dest: RegisterIndex(0), constant: ConstantIndex16(0) }
1: Closure { dest: RegisterIndex(1), proto: PrototypeIndex(0) }
2: SetUpTableCR { table: UpValueIndex(0), key: ConstantIndex8(1), value: RegisterIndex(1) }
3: GetUpTableC { dest: RegisterIndex(1), table: UpValueIndex(0), key: ConstantIndex8(2) }
4: GetUpTableC { dest: RegisterIndex(2), table: UpValueIndex(0), key: ConstantIndex8(1) }
5: LoadConstant { dest: RegisterIndex(3), constant: ConstantIndex16(3) }
6: Call { func: RegisterIndex(2), args: VarCount(Opt254(Some(1))), returns: VarCount(Opt254(None)) }
7: Call { func: RegisterIndex(1), args: VarCount(Opt254(None)), returns: VarCount(Opt254(Some(0))) }
8: Jump { offset: 0, close_upvalues: Opt254(Some(0)) }
9: Return { start: RegisterIndex(0), count: VarCount(Opt254(Some(0))) }
upvalues:
0: Environment
prototypes:
=============
FunctionProto(0x7ff69bc06be8)
=============
fixed_params: 1, has_varargs: false, stack_size: 2
opcodes:
0: GetUpValue { dest: RegisterIndex(1), source: UpValueIndex(0) }
1: AddRR { dest: RegisterIndex(1), left: RegisterIndex(1), right: RegisterIndex(0) }
2: Return { start: RegisterIndex(1), count: VarCount(Opt254(Some(1))) }
3: Return { start: RegisterIndex(0), count: VarCount(Opt254(Some(0))) }
upvalues:
0: ParentLocal(RegisterIndex(0))
``

### DIFF
--- a/src/bin/luster.rs
+++ b/src/bin/luster.rs
@@ -134,7 +134,6 @@ fn main() -> Result<(), Box<StdError>> {
         return Ok(());
     }
 
-    let file = io::buffered_read(File::open(matches.value_of("file").unwrap())?)?;
     lua.sequence(|root| {
         sequence::from_fn_with(root, |mc, root| {
             Ok(Closure::new(
--- a/src/bin/luster.rs
+++ b/src/bin/luster.rs
@@ -102,6 +102,12 @@ fn main() -> Result<(), Box<StdError>> {
                 .long("repl")
                 .help("Load into REPL after loading file, if any"),
         )
+        .arg(
+            Arg::with_name("dump")
+                .short("d")
+                .long("dump")
+                .help("Dump bytecode"),
+        )
         .arg(Arg::with_name("file").help("File to interpret").index(1))
         .get_matches();
 
@@ -114,6 +120,21 @@ fn main() -> Result<(), Box<StdError>> {
 
     let file = io::buffered_read(File::open(matches.value_of("file").unwrap())?)?;
 
+    if matches.is_present("dump") {
+        lua.sequence(|root| {
+            sequence::from_fn_with(root, |mc, root| {
+                let proto = compile(mc, root.interned_strings, file)?;
+                println!("{}", proto);
+                Ok(())
+            })
+            .map_ok(|_| ())
+            .map_err(|e: Error| e.to_static())
+            .boxed()
+        })?;
+        return Ok(());
+    }
+
+    let file = io::buffered_read(File::open(matches.value_of("file").unwrap())?)?;
     lua.sequence(|root| {
         sequence::from_fn_with(root, |mc, root| {
             Ok(Closure::new(
--- a/src/bin/luster.rs
+++ b/src/bin/luster.rs
@@ -106,6 +106,7 @@ fn main() -> Result<(), Box<StdError>> {
             Arg::with_name("dump")
                 .short("d")
                 .long("dump")
+                .conflicts_with("repl")
                 .help("Dump bytecode"),
         )
         .arg(Arg::with_name("file").help("File to interpret").index(1))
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by kyren** [comment]
> First of all, thanks for the amazing project! I learned a lot from it. :)

Thank you!

> For now, I just dump the FunctionProto with Debug trait with minimal formatting. It's definitely note a good textual representation of luster's bytecode. However, it's fine for learning purpose IMO.

It's already way better than https://github.com/kyren/luster/blob/master/src/bin/compiler.rs is, I had been meaning to fold that into the main luster binary for a while.

One thing that concerns me is that since Gc pointers can be recursive, the Display / Debug impls are slightly dangerous.  This isn't specific to your code (there was already a Debug impl there after all), but I do wonder what the right way to handle that is.

File: src/bin/luster.rs [hunk 1]
```diff
@@ -114,6 +120,21 @@ fn main() -> Result<(), Box<StdError>> {
...
+                Ok(())
+            })
+            .map_ok(|_| ())
+            .map_err(|e: Error| e.to_static())
+            .boxed()
+        })?;
+        return Ok(());
+    }
+
+    let file = io::buffered_read(File::open(matches.value_of("file").unwrap())?)?;
```
kyren: This is reading the file argument twice if not given the "dump" flag, maybe an `else` is appropriate here?  Also, it doesn't seem to complain if both `--dump` and `--repl` are given, even though it won't run the repl, though that might be fine for the moment.

poga: Thanks for the review!

I've removed the redundant read in https://github.com/kyren/luster/pull/14/commits/f92af5bba2616a5e61f77c928926b6055efc1116.

https://github.com/kyren/luster/pull/14/commits/e87af933231846a35ee803bac234c77d2ea73967 makes luster return an error if both `--dump` and `--repl` are given.

poga:
Oops, sorry I didn't notice https://github.com/kyren/luster/blob/master/src/bin/compiler.rs already exists. Maybe I should move the implementation to there instead of introducing a new flag?

kyren:
Actually, I did mention before that I wanted to fold the compiler into the main binary, but now I think I've changed my mind.  I think the compiler binary should be separate and then the parser / compiler functionality should be folded together.  The parser is really pretty dumb, but it's been useful in the past to debug parsing / compilation issues.  I guess move it to the compiler binary?

poga:
Done. https://github.com/kyren/luster/pull/14/commits/9578eec60dcc612314fb9d1bc2a94f68e83031a7

kyren:
Thank you very much for the PR!

PR #16, opened 2019-03-12
Add base function `assert`

g-plane:
Ref: https://www.lua.org/manual/5.3/manual.html#6.1


kyren:
Yep, that seems exactly right!

This should make writing tests slightly easier as well.

PR #20, opened 2019-05-01
Add string.len to the standard library

cjcole:


src/string.rs
```diff
@@ -99,7 +101,7 @@ impl<'gc> String<'gc> {
                 }
             }
         }
-        Ok(String::Long(Gc::allocate(mc, bytes.into_boxed_slice())))
+        Ok(String::new(mc, &bytes))
```
kyren: Yeah, that's much better good catch.

src/string.rs
```diff
@@ -110,6 +112,22 @@ impl<'gc> String<'gc> {
...
     }
+
+    pub fn len(&self) -> Result<i64, StringError> {
+        fn as_i64(len: usize) -> Result<i64, StringError> {
+            if len <= std::i64::MAX as usize {
+                Ok(len as i64)
+            } else {
+                Err(StringError::TooLong)
+            }
+        }
```
kyren: You should probably use num_traits::cast here, as it's what the rest of the crate does.

kyren: Ehhh, actually I take that back, probably this is an impossible error, so it should either be a panic or... just return i64::MAX or something.  Probably panic?  I mean how realistic is a string > i64::MAX?.

src/stdlib/string.rs
```diff
@@ -0,0 +1,34 @@
...
+                        )))
+                        .into()),
+                    }
+                }))
+            }),
+        )
+        .unwrap();
+
+    env.set(mc, String::new_static(b"string"), string).unwrap();
+}
```
kyren: This is perfectly correct, but I want to apologize for you having to write it.  This API is the biggest problem right now with luster, and probably the very, *very* next major thing to work on is callback args / returns to make this easier and faster.

src/stdlib/string.rs
```diff
@@ -0,0 +1,34 @@
...
+            mc,
+            String::new_static(b"len"),
+            Callback::new_sequence(mc, |args| {
+                Ok(sequence::from_fn_with(args, |mc, args| {
+                    match args
+                        .get(0)
+                        .cloned()
+                        .unwrap_or(Value::Nil)
+                        .to_string(mc)
+                        .and_then(|s| s.len().ok())
```
kyren: You can just use map here instead of `and_then(|v| <expr>.ok())`

src/stdlib/string.rs
cjcole: No worries, don't mind at all.

src/string.rs
cjcole: True, unlikely. I'll just panic.

src/stdlib/string.rs
```diff
@@ -0,0 +1,33 @@
...
+            Callback::new_sequence(mc, |args| {
+                Ok(sequence::from_fn_with(args, |mc, args| {
+                    match args
+                        .get(0)
+                        .cloned()
+                        .unwrap_or(Value::Nil)
+                        .to_string(mc)
+                    {
+                        Some(s) => Ok(CallbackResult::Return(vec![Value::Integer(s.len())])),
+                        _ => Err(RuntimeError(Value::String(String::new_static(
```
kyren: Really small thing, I think just None => is easier to read here.

cjcole: I'll fess up: that's a copy-paste from math.rs. But I agree and will update.

kyren: oh it's probably wrong there too :P

kyren:
Thanks a lot, this looks good, merging!

PR #21, opened 2019-05-03
Add string.{char, lower, upper, reverse} to stdlib

cjcole:
I believe that a "sufficiently smart compiler" would compile this String::new to the same machine code as the existing implementation. (Pretty sure rustc isn't that compiler, though.)

Could retain the existing implementation for performance.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -16,7 +16,6 @@ use crate::Value;
 #[collect(require_static)]
 pub enum StringError {
     Concat { bad_type: &'static str },
-    TooLong,
 }
 
 impl StdError for StringError {}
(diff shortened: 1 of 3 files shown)


src/string.rs
```diff
@@ -60,17 +58,32 @@ impl<'gc> Debug for String<'gc> {
 
 impl<'gc> String<'gc> {
     pub fn new(mc: MutationContext<'gc, '_>, s: &[u8]) -> String<'gc> {
+        String::new_mapped(mc, s, |from, idx| from[idx])
+    }
+
+    pub fn new_mapped<F>(mc: MutationContext<'gc, '_>, s: &[u8], f: F) -> String<'gc>
```
kyren: This is pretty specific and a bit over-kill for the current optimization level.  I'm aware that String construction is not the fastest, but what's really blocking it is dynamically sized types backed by a Gc pointer.  This is waiting on some details of the allocator system to be made stable.

An alternative would be to be able to construct a String from a `Box<[u8]>` for now, so if you need to allocate a `Vec<u8>` as an intermediary then it can just directly become that type of string.

Another alternative, maybe better, is to implement `FromIterator<u8>`.

XVilka:
Should be rebased to resolve conflicts

PR #24, opened 2019-05-10
Static string key

irxground:
The static string key is frequently written when creating a standard library, so I can write it simply.


kyren:
Sorry I disappeared for four years 😔 this is implemented now

PR #36, opened 2023-11-08
Added support for table.concat, table.insert, table.remove, tonumber, string.sub

svermeulen:
Also changed print() to use space instead of tab between arguments since this is how the official lua implementation seems to work.

Should also note that I'm new to Rust so please let me know if I'm not following best practices

### DIFF
--- a/src/stdlib/base.rs
+++ b/src/stdlib/base.rs
@@ -46,7 +46,9 @@ pub fn load_base<'gc>(ctx: Context<'gc>) {
 
                     if base.is_some() {
                         // TODO
-                        return Err("tonumber does not currently support base parameter".into_value(ctx).into())
+                        return Err("tonumber does not currently support base parameter"
+                            .into_value(ctx)
+                            .into());
                     }
 
                     stack.replace(ctx, v.to_number());
--- a/src/stdlib/base.rs
+++ b/src/stdlib/base.rs
@@ -37,6 +37,29 @@ pub fn load_base<'gc>(ctx: Context<'gc>) {
         .globals
         .set(
             ctx,
+            "tonumber",
+            AnyCallback::from_fn(&ctx, |ctx, _, stack| {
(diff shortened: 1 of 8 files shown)


kyren:
Thank you for the PR!

I haven't looked in detail yet at all, I wanted to reply to this comment first.

> Also changed print() to use space instead of tab between arguments since this is how the official lua implementation seems to work.

Lua 5.3 and Lua 5.4 both print tabs? That's specifically why I implemented it that way. What version of Lua are you comparing with?

kyren:
I haven't finished the review but I hit a snag.. I need to research *precisely* what `table.insert` and `table.remove` even do in PUC-Rio Lua first to even know what has to be done.

kyren:
I understand contributing to this project is no fun lol. That's actually pretty much the reason I stopped explicitly asking for contributions, there are too many things that are not so much work that has to be done but *questions that need to be answered*, lots of things are like this and just end up being a quagmire for a new contributor.

What would be great would be to make an official chat server, so that people who wanted to contribute could join and talk about this stuff synchronously, otherwise the back and forth over github PRs will be agonizingly slow.

If I made a discord server, would you be willing to talk there? Would matrix be better? I have a few other people who should probably join, but the last thing I really want to do is *manage* a discord server so I'm not sure what the best course of action is.

kyren:
I added a chat link to the README, a new #piccolo channel on the Spicy Lobster gamedev server, which I'll try to have all piccolo related discussions in for the foreseeable future. (Older discussions took place in #code)

svermeulen:
> Lua 5.3 and Lua 5.4 both print tabs? That's specifically why I implemented it that way. What version of Lua are you comparing with?

Ohh ok that makes sense.  I was comparing it to Lua 5.1.  I've reverted that change now.

I've joined the discord so will coordinate any work there going forward!  Thanks

PR #37, opened 2023-11-10
feat: make nesting more apparent in compiler example.

zicklag:
I found the compiler example useful for experimenting with.

This just adds a tiny update to indent nested prototypes so that it's easier to see the tree structure.


kyren:
Makes sense to me!

PR #41, opened 2024-04-16
use `#[doc(inline)]` for re-exports

jrobsonchase:
tbh, it feels like this should be the default for re-exported things. Rust doc search is pretty awful when they aren't inlined, since you end up seeing two items in the results, one of which is the re-export which requires yet another click to get to the page you're actually looking for.

Alternatively, the target of the re-export could be made non-public, which would have the same effect as far as docs go.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,6 +23,7 @@ pub mod types;
 pub mod userdata;
 pub mod value;
 
+#[doc(inline)]
 pub use self::{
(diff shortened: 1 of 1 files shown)


kyren:
I may change this back, I think the main problem was that almost none of the types were documented at all, and rustdoc behaves weirdly with re-exports *when the type has no documentation*.

Once every type is documented, I think removing it again will be cleaner.

PR #46, opened 2024-04-23
Add COMPATIBILITY.md

Jengamon:
Add a place to keep track of compatibility with PUC-Lua, to see what has been implemented, and
to mark what cannot/should not be implemented.

In repo instead of issue as it might be helpful to keep track of compatibility as the code changes.

PR #47, opened 2024-04-23
make math.log match PUC-Lua

Jengamon:
Matches the interface for piccolo's `math.log` function to that of PUC-Lua.

We take advantage of Rust's optimized (?) functions when the requested base matches one of the routines, but we can now handle any base directly.

### DIFF
--- a/src/stdlib/math.rs
+++ b/src/stdlib/math.rs
@@ -126,6 +126,7 @@ pub fn load_math<'gc>(ctx: Context<'gc>) {
             None => Some(v.ln()),
             Some(f) if f == 2.0 => Some(v.log2()),
             Some(f) if f == 10.0 => Some(v.log10()),
+            Some(f) if f == core::f64::consts::E => Some(v.ln()),
             Some(base) => Some(v.log(base)),
         }),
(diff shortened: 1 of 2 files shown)


Jengamon:
Drafting as this is a fix for stuff found in #46

src/stdlib/math.rs
```diff
@@ -119,13 +119,16 @@ pub fn load_math<'gc>(ctx: Context<'gc>) {
...
-        "log10",
-        callback("log10", &ctx, |_, v: f64| Some(v.log10())),
+        "log",
+        callback("log", &ctx, |_, (v, base): (f64, Option<f64>)| match base {
+            None => Some(v.ln()),
+            Some(f) if f == 2.0 => Some(v.log2()),
+            Some(f) if f == 10.0 => Some(v.log10()),
+            Some(f) if f == core::f64::consts::E => Some(v.ln()),
+            Some(base) => Some(v.log(base)),
+        }),
```
kyren, line 131: Don't handle each individual case, just use `v.log(base)` if the base is given.

PR #48, opened 2024-04-23
fix: math.random

Jengamon:
PUC-Lua has different behavior if `math.random(0)` is executed, as 0 is not in [1, n) for any n.
This changes the code to match that behavior.

Also fixes the crash from `math.random(v)` where `v` < 0 and makes it a Lua error.

I poked the function until it cried, so I included a fix for `math.random([n, ]math.maxinteger)`...


Jengamon:
Drafting as this is a fix for stuff found in https://github.com/kyren/piccolo/pull/46

kyren:
Yeah, this seems right to me, thank you!

PR #49, opened 2024-04-23
match math.randomseed to PUC-Lua

Jengamon:
~~The `math.randomseed(x, y)` impl might be a *little* extra.~~

Is no longer extra.

### DIFF
--- a/src/stdlib/math.rs
+++ b/src/stdlib/math.rs
@@ -236,7 +236,6 @@ pub fn load_math<'gc>(ctx: Context<'gc>) {
                             let mut seed = [0; 32];
                             let high_bytes = high.to_ne_bytes();
                             let low_bytes = low.to_ne_bytes();
-                            // Let's have some fun.
                             seed[..8].copy_from_slice(&low_bytes);
                             seed[8..16].copy_from_slice(&high_bytes);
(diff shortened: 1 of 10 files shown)


kyren:
> The math.randomseed(x, y) impl might be a little extra.

It's a little much, but I have absolutely no idea what the *best* thing to do would be, so I would just be inclined to pick something simple.

How about 32 bytes made up of [low,high,low,high]?

src/stdlib/math.rs
```diff
@@ -215,11 +215,39 @@ pub fn load_math<'gc>(ctx: Context<'gc>) {
...
+                    (Some(seed), None) | (Some(seed), Some(0)) => {
+                        *rng.borrow_mut() = SmallRng::seed_from_u64(seed);
+                        Some(())
+                    }
+                    (Some(high), Some(low)) => {
+                        let seed = {
+                            let mut seed = [0; 32];
+                            let high_bytes = high.to_ne_bytes();
+                            let low_bytes = low.to_ne_bytes();
+                            // Let's have some fun.
```
kyren, line 237: it's not fun anymore 😔

Jengamon:
git is a pain, will reopen in clean branch

PR #50, opened 2024-04-23
make type match PUC-Lua

Jengamon:
Use a little trick to tell the difference between `type(nil)` (ok and valid) and `type()` (bad and invalid).


kyren:
thank you!

PR #51, opened 2024-04-23
match select to PUC-Lua

Jengamon:


kyren:
This seems correct. I'm not wild about the intermediate `Vec` but I can change that, and the tests are good.

Jengamon:
> This seems correct. I'm not wild about the intermediate `Vec` but I can change that, and the tests are good.

Yeah sadly for select you need to know the size beforehand esp. in the case of negative indicies, as going *over* the size in the negative direction is *also* an error, so that `select(-1)` is an error (`select(-1, <anything>...)` is valid tho).

kyren:
It would have been more polite of me to properly review the PR and show you, but I had let these PRs sit for a while so I wasn't sure if you were around atm and I wanted to clear the backlog a bit.

You can use `Stack::drain` to drain elements out of the `Stack` just like `Vec::drain`, so you didn't need an intermediate `Vec`: 8b183bc8c93c0179fdaefd2e94551581c32ffc8b

kyren:
Also, you don't have to contort the tests to have functions and one big assert at the bottom, I think a lot of the older tests are weirdly structured because they predate `do end` blocks and basic constructs that exist and work now.

PR #52, opened 2024-04-23
Local attributes

Jengamon:
Right now, this only accepts the local attributes, but doesn't do anything with them.
As is, this would lead to incorrect behavior, as `const` and `close` change some of the mechanics
of how assignments work/are allowed, while any other values should be rejected.

## Goals
- [x] accept local attribute syntax
- [x] reject invalid attributes
- [x] implement `const` attribute
- [ ] implement `close` attribute (along with the `__close` metamethod)


Jengamon:
In the `tests/goldenfile.rs` you will see a module `collected_print` which is a clone of the print from the io implementation except it collects the output to throw it in a channel. I think this might raise the question of how to handle the io module.

Jengamon:
@Aeledfyr go ahead and make any changes as you see fit, right now i probably don’t have the time to finish it myself

Jengamon:
yea it probably makes more sense in the parser yeah

PR #57, opened 2024-04-27
Arithmetic metamethods

ZayadNimrod:
I found myself needing to override the addition operator (2d position+offset arithmetic). I've only implemented overriding `__add` and `__sub` so far; if this is how you'd like the other arithmetic metamethods implemented, I'll implement those too.

### DIFF
--- a/src/meta_ops.rs
+++ b/src/meta_ops.rs
@@ -442,101 +442,44 @@ pub fn arithmetic_meta_op<'gc>(
         (Value::Number(a), Value::Number(b)) => (raw_op(Value::Number(a), Value::Number(b)))
             .map(MetaResult::Value)
             .or(Err(()))?,
-
-        (Value::Table(a), b) => {
-            if let Some(metamethod) = get_metamethod_from_table(ctx, a, m) {
-                MetaResult::Call(MetaCall {
-                    function: call(ctx, metamethod).or(Err(()))?,
-                    args: [a.into(), b.into()],
-                })
-            } else {
-                match b {
-                    Value::Table(b) => {
-                        if let Some(b_eq) = get_metamethod_from_table(ctx, b, m) {
-                            MetaResult::Call(MetaCall {
-                                function: call(ctx, b_eq).or(Err(()))?,
-                                args: [a.into(), b.into()],
-                            })
-                        } else {
-                            Err(())?
-                        }
-                    }
-                    Value::UserData(b) => {
-                        if let Some(b_eq) = get_metamethod_from_userdata(ctx, b, m) {
-                            MetaResult::Call(MetaCall {
-                                function: call(ctx, b_eq).or(Err(()))?,
-                                args: [a.into(), b.into()],
-                            })
-                        } else {
-                            Err(())?
-                        }
-                    }
-                    Value::Nil => todo!(),
-                    _ => Err(())?,
-                }
-            }
+        // Could avoid calling `get_metamethod_from_{table,userdata}` twice if if-let-match gets stabilised
+        (Value::Table(a), b) if get_metamethod_from_table(ctx, a, m).is_some() => {
+            let metamethod = get_metamethod_from_table(ctx, a, m).unwrap();
+            MetaResult::Call(MetaCall {
+                function: call(ctx, metamethod).or(Err(()))?,
+                args: [a.into(), b],
+            })
         }
 
-        (Value::UserData(a), b) => {
-            if let Some(metamethod) = get_metamethod_from_userdata(ctx, a, m) {
-                MetaResult::Call(MetaCall {
-                    function: call(ctx, metamethod).or(Err(()))?,
-                    args: [a.into(), b.into()],
-                })
-            } else {
-                match b {
-                    Value::Table(b) => {
-                        if let Some(b_eq) = get_metamethod_from_table(ctx, b, m) {
-                            MetaResult::Call(MetaCall {
-                                function: call(ctx, b_eq).or(Err(()))?,
-                                args: [a.into(), b.into()],
-                            })
-                        } else {
-                            Err(())?
-                        }
-                    }
-                    Value::UserData(b) => {
-                        if let Some(b_eq) = get_metamethod_from_userdata(ctx, b, m) {
-                            MetaResult::Call(MetaCall {
-                                function: call(ctx, b_eq).or(Err(()))?,
-                                args: [a.into(), b.into()],
-                            })
-                        } else {
-                            Err(())?
-                        }
-                    }
-                    Value::Nil => todo!(),
-                    _ => Err(())?,
-                }
-            }
+        (Value::UserData(a), b) if get_metamethod_from_userdata(ctx, a, m).is_some() => {
+            let metamethod = get_metamethod_from_userdata(ctx, a, m).unwrap();
+            MetaResult::Call(MetaCall {
+                function: call(ctx, metamethod).or(Err(()))?,
+                args: [a.into(), b],
+            })
         }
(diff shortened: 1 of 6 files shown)


src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+                })
+            } else if let Some(b_eq) = get_sub(b) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, b_eq)?,
+                    args: [a.into(), b.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "tables with no __sub",
```
Jengamon, line 479: If generic over metamethod, this should reflect the metamethod used, so probably a `format!("tables with no {m}")`

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+                })
+            } else if let Some(b_eq) = get_sub(b) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, b_eq)?,
+                    args: [a.into(), b.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "userdata(s) with no __sub",
```
Jengamon, line 498: ditto

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+                })
+            } else if let Some(b_eq) = get_metamethod_from_table(ctx, b, m) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, b_eq)?,
+                    args: [a.into(), b.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "userdata(s) with no __sub",
```
Jengamon, line 516: ditto

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+                })
+            } else if let Some(b_eq) = get_metamethod_from_userdata(ctx, b, m) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, b_eq)?,
+                    args: [a.into(), b.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "userdata(s) with no )__sub",
```
Jengamon, line 534: ditto

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+            };
+            if let Some(a_eq) = get_sub(t) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, a_eq)?,
+                    args: [t.into(), a.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "table with no __sub",
```
Jengamon, line 559: ditto

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+            };
+            if let Some(a_eq) = get_sub(t) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, a_eq)?,
+                    args: [t.into(), a.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "userdata with no __sub",
```
Jengamon, line 583: ditto

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+            };
+            if let Some(a_eq) = get_sub(t) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, a_eq)?,
+                    args: [a.into(), t.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "table with no __sub",
```
Jengamon, line 608: ditto

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+            let get_sub = |t| get_metamethod_from_userdata(ctx, t, MetaMethod::Sub);
+            if let Some(a_eq) = get_sub(t) {
+                MetaResult::Call(MetaCall {
+                    function: call(ctx, a_eq)?,
+                    args: [a.into(), t.into()],
+                })
+            } else {
+                Err(TypeError {
+                    expected,
+                    found: "userdata with no __sub",
```
Jengamon, line 622: ditto

src/meta_ops.rs
```diff
@@ -396,3 +400,296 @@ pub fn equal<'gc>(
...
+                })?
+            }
+        }
+
+        (Value::Nil, _) => Err(TypeError {
```
Jengamon, line 627: these could probably be merged into 1 case, that uses `Value::type_name` (iirc) for the found typename.

Jengamon (commented):
just something i noticed as a review to group all the comments together.

ZayadNimrod:
This should address the suggestions re: error messages.  That said, this requires compile-time construction of 
those strings (using `format!` means constructing a string every time an error is raised, and to make those the `&str` that `TypeError` takes would require leaking the constructed `String`), so I've brought in a macro crate to handle that. Not sure if the resulting code is particularly nice to work with, and the extra dependency is probably undesirable.

Jengamon (commented):
I'd say try not to bring in external dependencies, so don't actually return a `Result` from your helper `arithmetic_meta_op`, instead going for an `Option<_>`, and when returning from the meta method, generating the error at that point

ZayadNimrod:
Not quite comfortable with erasing error sources like this so I'm keeping the commit in until a last-minute history squash.

Jengamon:
> Not quite comfortable with erasing error sources like this so I'm keeping the commit in until a last-minute history squash.

Note that I'm not telling you to do X or Y, so if you're not comfortable with erasing error sources, don't! Find a way to get your nice errors. Just be aware that I am *not* operating under any official capacity. My suggestions don't really have binding force, or any force at all. You should do you. I was just questioning the necessity of bringing in an entire other crate for the const evaluation of formatted string when we could just not, and figure out a way to generate w/e error we want with w/e data we want.

src/meta_ops.rs
```diff
@@ -396,3 +400,222 @@ pub fn equal<'gc>(
...
+            } else {
+                Err(())?
+            }
+        }
+
+        (Value::Table(t), a) => {
+            let get_sub = |t: Table<'gc>| {
+                let eq = t
+                    .metatable()
+                    .map(|t| t.get(ctx, MetaMethod::Sub))
```
kyren, line 520: I think there are several places where MetaMethod::Sub is still used instead of the metamethod parameter?

ZayadNimrod: Fixed now. I've condensed the code down further so there shouldn't be any more cases.

ZayadNimrod:
Unsure if bc144f0d412db2da66c31f05936216a16504afa7 should make it in, since it should result in extra lookups to the metatable. But it does make the code more compact and less duplicated. Either way this will need a bunch of history surgery once the code is settled.

ZayadNimrod:
Sorry to ping again, but it's been a month, any thoughts on this before I add the remainder of the arithmetic methods?

kyren:
I'm very sorry for dropping the ball on this PR, I should have been more responsive. 😔

This has been implemented in #74 though, so we have a *very nearly* complete set of implemented metamethods now!

Sorry you had to do all this work for a PR that was never merged, but I still very much appreciate the effort you spent, thank you. ❤️

PR #59, opened 2024-04-28
string.char implementation

Jengamon:
I'm not 100% on the error messages, so i modeled them after PUC-Lua's.


Jengamon:
This would actually work better as an implementation of `utf8.char`

Jengamon:
right now the implementation charges 1 fuel per argument, but can/should be adjusted?

PR #60, opened 2024-04-28
initial impl of string.lower,reverse,upper and clippy fixes for string.sub

Jengamon:
I have initial sketch of `string.lower`, ~~`string.rep`~~, `string.reverse`, and `string.upper`.


Jengamon:
These would work "logically" under the `utf8` table, but PUC-Lua has none of these functions there, so it's an open question as to where these should go.

Jengamon:
`string.rep` is also charging 1 fuel per repetition.

kyren:
thank you!

PR #63, opened 2024-05-07
Add COMPATIBILITY.md

Jengamon:
Fresh based off master so that only the document is added

Supersedes #46


kyren:
The basic document is good, and I want  to merge this whenever the PR is not marked as draft, but *be forewarned* that I am going to edit this with impunity.

PR #66, opened 2024-05-13
fix: allow seed to be any size

Jengamon:
Fix the seed code so that instead of being fixed to [u8; 32], the `SeedableRng::Seed` type is allowed to be `[u32; N]` (through the use of `std::array::from_fn`), and the seed is simply the low and high bytes repeated in order, ad infinitum.


kyren:
Ah yeah, good catch!

PR #67, opened 2024-05-14
implement string.sub

Jengamon:
I wanted to sub my strings, so I made a quick implementation


Jengamon:
This is related/dup-implementation of part of #36, however this handles strings as byte arrays, which corresponds with PUC-Rio Lua behavior

PR #70, opened 2024-05-15
Mirror `__index` change to `__newindex`

Jengamon:
realized this should be the case from Lua 5.4 manual:

>Like with indexing, the metavalue for this event can be either a function, a table, or any value with an __newindex metavalue. If it is a function, it is called with table, key, and value as arguments. Otherwise, Lua repeats the indexing assignment over this metavalue with the same key and value. This assignment is regular, not raw, and therefore can trigger another __newindex metavalue.


kyren:
yeah that makes sense

PR #75, opened 2024-06-21
feat: implement PUC-Rio tonumber

Jengamon:
Implements `tonumber` according to the Lua 5.4 manual https://www.lua.org/manual/5.4/manual.html#pdf-tonumber.


src/stdlib/base.rs
```diff
@@ -8,6 +8,76 @@ use crate::{
...
+        Callback::from_fn(&ctx, |ctx, _, mut stack| {
+            fn extract_number_data(bytes: &[u8]) -> (&[u8], i64) {
+                let first_nonspace = bytes.iter().position(|b| !b.is_ascii_whitespace());
+                let last_nonspace = bytes.iter().rposition(|b| !b.is_ascii_whitespace());
+                let bytes = match (first_nonspace, last_nonspace) {
+                    (Some(f), Some(l)) => &bytes[f..=l],
+                    (Some(f), None) => &bytes[f..],
+                    (None, Some(l)) => &bytes[..=l],
+                    (None, None) => bytes,
+                };
```
kyren, line 22: You can use trim_whitespace from the lexer here

src/stdlib/base.rs
```diff
@@ -8,6 +8,76 @@ use crate::{
...
+
+                let is_neg = bytes.first().is_some_and(|b| *b == b'-');
+                let bytes = if is_neg || bytes.first().is_some_and(|b| *b == b'+') {
+                    &bytes[1..]
+                } else {
+                    bytes
+                };
+                let sign = if is_neg { -1 } else { 1 };
+                (bytes, sign)
```
kyren, line 31: This is the same as `read_neg` from the lexer which can be made public so it can be used here as well.

src/stdlib/base.rs
```diff
@@ -8,6 +8,76 @@ use crate::{
...
+                    // represent an integer of the base
+                    const BASE_ELEMENTS: &[u8; 36] = b"0123456789abcdefghijklmnopqrstuvwxyz";
+                    let byte_values = bytes
+                        .iter()
+                        .map(|b| {
+                            BASE_ELEMENTS
+                                .iter()
+                                .position(|i| b.to_ascii_lowercase() == *i)
+                        })
+                        .collect::<Vec<_>>();
```
kyren, line 56: I think you can avoid collecting into a Vec here because the resulting iterator can be cloned?

src/stdlib/base.rs
```diff
@@ -8,6 +8,76 @@ use crate::{
...
+                    }
+                    // Since a base [2 - 36 are the only valid values] was specified, this *must*
+                    // represent an integer of the base
+                    const BASE_ELEMENTS: &[u8; 36] = b"0123456789abcdefghijklmnopqrstuvwxyz";
+                    let byte_values = bytes
+                        .iter()
+                        .map(|b| {
+                            BASE_ELEMENTS
+                                .iter()
+                                .position(|i| b.to_ascii_lowercase() == *i)
```
kyren, line 54: The whole BASE_ELEMENTS array thing is much slower than the alternative of checking whether the character is in the range '0'-'9' and subtracting from '0', or checking whether the character is in the range 'a' through 'z' and subtracting 'a' and adding 10.

src/stdlib/base.rs
```diff
@@ -8,6 +8,76 @@ use crate::{
...
+                        .iter()
+                        .any(|v| v.is_none() || v.is_some_and(|v| v >= base as usize))
+                    {
+                        Value::Nil
+                    } else {
+                        let mut value: i64 = 0;
+                        for (idx, elem) in byte_values.into_iter().rev().enumerate() {
+                            let Some(elem) = elem else {
+                                // The earlier byte value check means that this should not be hit
+                                unreachable!();
```
kyren, line 67: it's probably fine if you just want to `return Ok(Value::Nil)` here instead of doing a pre-check pass, which might be a lot simpler.

kyren:
Looks good now, thank you!

PR #76, opened 2024-06-22
match `pairs` behavior to PUC-Rio

Jengamon:
This should fix the COMPATIBILITY note.


src/stdlib/base.rs
```diff
@@ -212,13 +212,33 @@ pub fn load_base<'gc>(ctx: Context<'gc>) {
...
+                    fn poll(
+                        &mut self,
+                        _ctx: Context<'gc>,
+                        _exec: Execution<'gc, '_>,
+                        mut stack: Stack<'gc, '_>,
+                    ) -> Result<SequencePoll<'gc>, Error<'gc>> {
+                        // QUESTION how do you just consume the data w/o doing anything?
+                        if stack.len() > 3 {
+                            _ = stack.drain(3..).count();
+                        }
```
Aeledfyr, line 230: The `drain` iterator will remove the values when dropped, even if the iterator is not consumed, so you can just do `stack.drain(3..)`

Jengamon: Awesome thanks!

PR #80, opened 2024-06-24
Unify string conversion logic, remove Value::to_string footgun

Aeledfyr:
This makes `Value` no longer directly implement `fmt::Display`, so it no longer has a `to_string` method.  This is probably a good thing, as calling `to_string` would silently corrupt non-utf8 data.

Method changes:
- rename `Value::display` to `Value::write`
- add `Value::display(self) -> impl Display`
- add `Value::into_string(self, ctx) -> Option<String<'gc>>`
- add `Value::is_implicit_string(self) -> bool`

Adds a `FromValue` impl for `string::String<'gc>`

This significantly reduces the boilerplate for stdlib methods that deal with strings.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -286,7 +286,6 @@ macro_rules! impl_from {
 }
 impl_from! {
     [Boolean bool],
-    [String String<'gc>],
     [Table Table<'gc>],
     [Function Function<'gc>],
     [Thread Thread<'gc>],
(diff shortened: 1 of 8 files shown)


src/stdlib/strfmt.c
Jengamon, line 1: you might have left this in by accident?

src/stdlib/strlib.c
Jengamon, line 1: Also this

src/stdlib/strmatch.c
Jengamon, line 1: and this

src/stdlib/strpack.c
Jengamon, line 1: And this.

Jengamon (commented):
some C files were left in

kyren:
I really like this PR!

This changes the semantics of `impl<'gc> FromValue<'gc> for String<'gc> {...}` to also convert numbers / integers to strings automatically.

This is a good change, because it makes the behavior of FromValue for String mirror the behavior of FromValue for i64 / f64... the numeric FromValue impls convert *from* a string, so the String FromValue impl should convert *from* numbers.

I have already begun to regret the auto-conversions to/from String. I know that this is the semantics of Lua itself, but even PUC-Rio Lua has started to back off of it with the eventual goal of only using metatables to support conversion, so that it can be turned *off*.

I have vague plans to support turning off this kind of auto conversion either by doing the conversion through metatables or through whatever BYOV (Bring Your Own Value) turns into, but these are pretty far off. In the meantime, this is strictly more consistent, and simpler.

PR #84, opened 2024-06-25
Implement async sequences

kyren:


src/async_callback.rs
```diff
@@ -0,0 +1,577 @@
...
+    ///
+    /// Ending an async block *without* calling `SequenceState::return_to` will instead *drop* the
+    /// values in the stack.
+    pub fn return_to(self) {
+        visit_shared(move |shared| {
+            *shared.next = SequencePoll::Return;
+        });
+    }
+
+    /// Call the given Lua function with arguemnts / returns starting at `bottom` in the Stack.
```
Aeledfyr, line 336: Minor spelling issue (arguments)

src/async_callback.rs
```diff
@@ -0,0 +1,577 @@
...
+            if let Some(err) = shared.error.take() {
+                Err(Local::stash(&shared.ctx, shared.locals, err))
+            } else {
+                Ok(())
+            }
+        })
+    }
+
+    /// Finish the async sequence by calling the given Lua function with the arguments as the
+    /// current contents of the stack.
```
Aeledfyr, line 359: Maybe "by calling the given Lua function with the current contents of the stack as arguments."?

src/async_callback.rs
```diff
@@ -0,0 +1,577 @@
...
+        visit_shared(move |shared| {
+            if let Some(err) = shared.error.take() {
+                Err(Local::stash(&shared.ctx, shared.locals, err))
+            } else {
+                Ok(())
+            }
+        })
+    }
+
+    /// Finish the async sequence by yeilding the current contents of the stack to the caller (or
```
Aeledfyr, line 391: Minor spelling issue (yielding)

src/async_callback.rs
```diff
@@ -0,0 +1,577 @@
...
+            *shared.next = SequencePoll::TailResume {
+                thread: thread.fetch(shared.locals),
+            };
+        });
+    }
+}
+
+/// A collection of stashed values that are local to a specific `AsyncSequence`.
+///
+/// `Local` values are branded by `'seq` and cannot escape their parent `AsyncSequence` and are (for
```
Aeledfyr, line 437: You may want to add [reference links](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html) for some of these to make the docs easier to navigate.  ``[`Local`]`` should be parsed as a link to the `Local` struct, since it's in-scope; in general, normal Rust path syntax should work, and `cargo doc` will warn on unresolved links.

src/async_callback.rs
```diff
@@ -0,0 +1,577 @@
...
+/// the purposes of garbage collection) considered *owned* by the parent `AsyncSequence`. Because of
+/// this, they correctly mimic what we could do if async blocks themselves could be traced, and so
+/// can't lead to uncollectable cycles with their parent.
+///
+/// NOTE: Calling `mem::forget` on a `Local` or creating cylces withing a single set of `Local`
```
Aeledfyr, line 442: Minor spelling issue (cycles)

src/registry.rs
```diff
@@ -251,129 +70,36 @@ impl<'gc> Registry<'gc> {
...
     ///
     /// It can be implemented for external types by implementing the `Stashable` trait.
+    ///
+    /// Values stashed in the global registry are not designed to be held within the Lua state!
+    /// They are 'static handles, not garbage collected values, and thus they are not *traced* like
+    /// garbage collected values to enable cycle collection.
+    ///
+    /// Any value stashed in the global registry will never be freed as long as:
+    ///   1) The global registry itself is alive, and...
+    ///   2) The returned stashed handle is not dropped.
```
Aeledfyr, line 85: I'm not sure how to phrase this better, but it's not immediately clear from this that dropping the stashed handle will lead to the value being eventually freed, even if the registry is still alive.  Maybe something along the lines of "[a stashed value] will never be freed as long as both the global registry and the stashed handle are alive."?
(Removing the double negative on the stashed handle)

Aeledfyr (approved):
Minor comments on documentation, but otherwise this looks good.

The only other question is whether having the async sequences return something equivalent to `CallbackReturn` would lead to a cleaner API or implementation -- I'm not sure, but it seems like could simplify some of the internal logic for polling the future.

kyren:
> The only other question is whether having the async sequences return something equivalent to CallbackReturn would lead to a cleaner API or implementation -- I'm not sure, but it seems like could simplify some of the internal logic for polling the future. (The current approach is fine, though, and changing it later wouldn't require significant changes to user code.)

We could do this... really I just wanted to stem the proliferation of return types.

I think the internal logic *after* it was changed to panic under misuse is not so bad?

PR #87, opened 2024-06-30
Implement `table.insert` and `table.remove`

Aeledfyr:
- Implement `table.insert` and `table.remove`, supporting metamethods
- For plain tables (no index/newindex/len metamethods), optimize inserts and removes that work entirely within the array portion of the table; this is the normal use-case of `table.insert` and `table.remove`, so it should be worth the minor added complexity.

This now includes a reasonable set of tests, though the edge cases of the optimized case may be difficult to trigger explicitly.

### DIFF
--- a/src/stdlib/table.rs
+++ b/src/stdlib/table.rs
@@ -5,6 +5,7 @@ use gc_arena::Collect;
 use crate::async_sequence::{
     AsyncSequence, LocalError, LocalFunction, LocalTable, LocalValue, Locals,
 };
+use crate::fuel::count_fuel;
 use crate::meta_ops::{self, MetaResult};
 use crate::table::RawArrayOpResult;
 use crate::{
@@ -126,13 +127,15 @@ pub fn load_table<'gc>(ctx: Context<'gc>) {
         Ok(())
     }
 
+    const FUEL_PER_SHIFTED_ITEM: i32 = 1;
+
     // Compat note with PRLua:
     // When the table is empty, table.remove(t, #t) will return nil
     // (even if the table has a 0 element).
     table.set_field(
         ctx,
         "remove",
-        Callback::from_fn(&ctx, |ctx, _, mut stack| {
+        Callback::from_fn(&ctx, |ctx, mut exec, mut stack| {
             let (table, index): (Table, Option<i64>) = stack.consume(ctx)?;
             let length;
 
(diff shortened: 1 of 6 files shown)


src/stdlib/table.rs
```diff
@@ -56,9 +62,323 @@ pub fn load_table<'gc>(ctx: Context<'gc>) {
...
+) -> Result<(), LocalError<'seq>> {
+    let call = seq.try_enter(|ctx, locals, _, stack| {
+        let table = locals.fetch(table);
+        let call = meta_ops::index(ctx, Value::Table(table), Value::Integer(key))?;
+        Ok(prep_metaop_call(ctx, stack, locals, call))
+    })?;
+    if let Some(call) = call {
+        seq.call(&call, bottom).await?;
+        seq.enter(|_, _, _, mut stack| {
+            stack.resize(bottom + 1); // Truncate stack
```
kyren, line 104: This is of course correct, but it makes me wonder if we shouldn't do this *within* meta_ops implementations, like always truncate to the expected number of returns for that specific metacall.

Not something that needs to change now, just thinking out loud really.

kyren:
This is *really* good work, thank you so much!

We talked about this a bit in chat, but there's some API changes I want to make to the raw table portion. Mostly, I don't want RawTable to have an API which only makes sense to use as part of the stdlib, so I'd like to change it to be more useful generally.. I'd like it to go ahead and correctly insert / remove and shift even when there are indexes in the map portion of the table.

This change is very good though, so I want to go ahead and merge it, and those changes can be done in a separate commit.

Thank you again!

src/stdlib/table.rs
Aeledfyr: It should be relatively simple, but right now the `Sequence` call logic skips a lot of the stack management of normal function calls; `ThreadState::return_to` currently does nothing when returning to a `Sequence`.  (When returning to a Lua frame, it truncates the stack.)

kyren: Just to be clear, that's how it's designed right now so it's not *skipped*, it just has nothing to do (but maybe you already understood that).

We COULD make the stack protocol for callbacks more featureful and also allow them to specify a stack truncation like Lua frames, that might be simpler, but it does complicate the CallbackReturn API a little.

What I was actually suggesting was not that at all, but returning a callback that did the truncation manually. If we have a lot of places where we want to do this though, it might make sense to make it part of the stack protocol for callbacks.

Is that what you were suggesting?

Aeledfyr: > returning a callback that did the truncation manually

If by this you mean wrapping the metacall `Function` in a Rust-implemented `Callback` that truncates the stack, it seems like a lot of unnecessary overhead?  In general, I think the async helper function is a reasonable place to put the stack management, unless we add explicit truncation to returns to `Sequence`s, which doesn't seem necessary at this point.

(That was what I was suggesting, but I also don't think it's worth the complexity at this point).

In this case, it may even have been better to pop the value off of the stack, and return it as a `LocalValue`, keeping the stack management entirely within `index_helper`.

PR #88, opened 2024-07-03
Implement `__concat`, `table.concat`, and add Lua polyfills for `table.sort` and `table.move`

Aeledfyr:
At some point it would be good to have `table.move` and `table.sort` implemented in Rust, but for now a Lua implementation works.

This PR also implements the `__concat` metamethod, and uses it to implement `table.concat`; unlike PRLua's implementation, this version supports tables with the `__concat` metamethod.

For example, `table.concat({ a, b, c }, sep)` is equivalent to `a .. sep .. b .. sep .. c`, and respects the right-associativity of the `..` operator.

Semantics questions on `table.concat`:
- Should `table.concat({ x }) == x`, regardless of type?  It currently does, as it never tries to call the `__concat` metaoperator.
- Is the complexity of this worth it over only supporting string/integers in `table.concat`?

### DIFF
--- a/src/meta_ops.rs
+++ b/src/meta_ops.rs
@@ -154,6 +154,8 @@ pub enum MetaOperatorError {
     Binary(MetaMethod, &'static str, &'static str),
     #[error("invalid table key")]
     IndexKeyError(#[from] InvalidTableKey),
+    #[error("concatenation result is too long")]
+    ConcatOverflow,
 }
 
 #[derive(Debug, Copy, Clone, Error)]
@@ -752,23 +754,37 @@ pub enum ConcatMetaResult<'gc> {
     Call(Function<'gc>),
 }
 
+/// Returns an estimate of the length of the concatenation of a list of values,
+/// or returns [`None`] if any value is not implicitly coercible to a string.
+pub fn estimate_concatenated_len<'gc>(
+    values: &[Value<'gc>],
+) -> Result<Option<usize>, MetaOperatorError> {
+    let mut len = 0usize;
+    for value in values {
+        let value_len = match value {
+            Value::Integer(i) => i.ilog10() as usize + i.is_negative() as usize,
+            Value::Number(_n) => 10,
+            Value::String(s) => s.as_bytes().len(),
+            _ => return Ok(None),
+        };
+        len = len
+            .checked_add(value_len)
+            .ok_or(MetaOperatorError::ConcatOverflow)?;
+    }
+    Ok(Some(len))
+}
+
 pub fn concat_many<'gc>(
     ctx: Context<'gc>,
     values: &[Value<'gc>],
 ) -> Result<ConcatMetaResult<'gc>, MetaOperatorError> {
-    // Since we have to make two passes, might as well estimate the length
-    let mut len = 0;
-    for value in values {
-        match value {
-            Value::Integer(i) => len += i.ilog10() as usize + i.is_negative() as usize,
-            Value::Number(_n) => len += 10,
-            Value::String(s) => len += s.as_bytes().len(),
-            _ => {
-                let func = Callback::from_fn(&ctx, concat_impl).into();
-                return Ok(ConcatMetaResult::Call(func));
-            }
-        }
-    }
+    // Since we have to make two passes to check for complex types,
+    // estimate the length in the first pass.
+    let Some(len) = estimate_concatenated_len(values)? else {
+        // Fall back to a sequence-based implemenation to handle metamethods
+        let func = Callback::from_fn(&ctx, concat_impl).into();
+        return Ok(ConcatMetaResult::Call(func));
+    };
 
     let mut bytes = Vec::with_capacity(len);
     for value in values {
@@ -793,9 +809,8 @@ fn concat_impl<'gc>(
             let call = seq.try_enter(|ctx, locals, _, mut stack| {
                 let bottom = i - 1;
                 let call = concat(ctx, stack[i - 1], stack[i])?;
-                Ok(prepare_async_metaop(
-                    ctx, &mut stack, locals, bottom, call, 1,
-                ))
+                let p = prepare_async_metaop(ctx, &mut stack, locals, bottom, call, 1);
+                Ok(p)
             })?;
             call.execute(&mut seq).await?;
         }
--- a/src/meta_ops.rs
+++ b/src/meta_ops.rs
@@ -4,10 +4,9 @@ use gc_arena::Collect;
 use thiserror::Error;
 
 use crate::async_callback::prepare_async_metaop;
-use crate::{async_sequence, Execution, SequenceReturn, Stack};
+use crate::{async_sequence, SequenceReturn};
 use crate::{
-    table::InvalidTableKey, Callback, CallbackReturn, Context, Error, Function, IntoValue, Table,
-    Value,
+    table::InvalidTableKey, Callback, CallbackReturn, Context, Function, IntoValue, Table, Value,
 };
 
 /// An enum of every possible Lua metamethod.
--- a/src/meta_ops.rs
+++ b/src/meta_ops.rs
@@ -3,6 +3,7 @@ use std::io::Write;
 use gc_arena::Collect;
 use thiserror::Error;
 
+use crate::async_callback::prepare_async_metaop;
 use crate::{async_sequence, Execution, SequenceReturn, Stack};
 use crate::{
     table::InvalidTableKey, Callback, CallbackReturn, Context, Error, Function, IntoValue, Table,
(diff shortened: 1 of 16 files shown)


kyren:
This is really good, thank you!

> Should table.concat({ x }) == x, regardless of type? It currently does, as it never tries to call the __concat metaoperator.

I think so, I don't think that the `__concat` metamethod should be called with a single argument definitely, and it would be weirder to call something like `__tostring`. I think it makes sense because it makes `table.concat({ table.concat({ x }), y})` the same as `table.concat({x, y})`.

> Is the complexity of this worth it over only supporting string/integers in table.concat?

This I'm not as sure about, but it makes complete sense to me to try to use the same implementation of the concat metaop for table.concat

src/meta_ops.rs
```diff
@@ -735,7 +736,113 @@ pub fn less_equal<'gc>(
...
     })
 }
+
+#[derive(Debug, Clone, Collect)]
+#[collect(no_drop)]
+pub struct ConcatMetaCall<'gc> {
+    pub function: Function<'gc>,
+    // Needs to be owned to work around borrowing issues in
+    // LuaFrame::call_meta_function
+    pub args: Vec<Value<'gc>>,
```
kyren, line 750: I have a plan for something else to do here. The borrowing issue is real because it would reference a slice in a Lua frame stack and call a function that is expected to manipulate the stack.

Instead, we just shouldn't use `LuaFrame::call_meta_function` and make one specific for concat. Instead of return arguments here, or even having a `ConcatMetaCall` struct, `ConcatMetaResult` should just return a function, and the documentation should say that the function should be called on the `values` provided to `concat`.

src/meta_ops.rs
```diff
@@ -735,7 +736,113 @@ pub fn less_equal<'gc>(
...
+#[collect(no_drop)]
+pub enum ConcatMetaResult<'gc> {
+    Value(Value<'gc>),
+    Call(ConcatMetaCall<'gc>),
+}
+
+pub fn concat<'gc>(
+    ctx: Context<'gc>,
+    values: &[Value<'gc>],
+) -> Result<ConcatMetaResult<'gc>, MetaOperatorError> {
```
kyren, line 763: This is a total bikeshed but personally I prefer `concat_single` -> `concat` and `concat` -> `concat_many`, making the two argument concat the "fundamental" one.

src/stdlib/table.rs
```diff
@@ -731,3 +759,130 @@ fn array_insert_shift<'gc>(
...
     (inner(table, length, key, value), length)
 }
+
+#[derive(Debug, Copy, Clone, Error)]
+pub enum TableConcatError {
+    #[error("resulting string too long in table.concat")]
+    Overflow,
+}
+
+fn concat_impl<'gc>(ctx: Context<'gc>, sep: Option<Value<'gc>>) -> BoxSequence<'gc> {
```
kyren, line 769: This has a lot of repeated logic with `meta_ops::concat`. I might be missing something, but are they identical except for the inclusion of an optional separator (and start / end index?)

This may sound heretical, but what if we implemented this by *interspersing* the separator into the stack and then calling `meta_ops::concat` so as not to duplicate everything?

src/stdlib/table.rs
```diff
@@ -731,3 +759,130 @@ fn array_insert_shift<'gc>(
...
+                let sep_len = sep_str.map(|s| s.len() as usize).unwrap_or(0);
+                let total_len = stack
+                    .len()
+                    .saturating_sub(1)
+                    .checked_mul(sep_len)
+                    .ok_or(TableConcatError::Overflow)?
+                    .checked_add(len)
+                    .ok_or(TableConcatError::Overflow)?;
+
+                // Should this be allocated in-place in the GC heap?
```
kyren, line 804: Unfortunately this is impossible right now, this is an open design question in `gc-arena`. `gc-arena` currently relies on size being a static property of an allocated type. There's a hack (partially) around this in piccolo's `String` but it's horrible.

kyren:
We're accumulating a lot of stdlib functions which *sort of* have unbounded time usage, but that unbounded time usage is proportional to some allocated memory (number of arguments on the stack, size of a table, etc).

I'm not fussed about solving this *now*, especially varargs because I think there might need to be a more centralized solution that limits varargs to something reasonable (I would be fine with a lower limit than PUC-Rio's limit even, which is something silly like a million). I'm mostly pointing this out to LET IT BE KNOWN that I am aware of it, and that it is a temporary problem.

src/stdlib/table.rs
```diff
@@ -731,3 +759,130 @@ fn array_insert_shift<'gc>(
...
+                let mut len = 0;
+                // Since we have to make two passes, might as well estimate the length
+                for value in &stack[..] {
+                    match value {
+                        Value::Integer(i) => len += i.ilog10() as usize + i.is_negative() as usize,
+                        Value::Number(_n) => len += 10,
+                        Value::String(s) => len += s.as_bytes().len(),
+                        _ => return Ok(stack.len()),
+                    }
+                }
```
kyren, line 793: This can also overflow?

src/stdlib/table.rs
Aeledfyr: The start/end index is handled by `table.unpack`, so besides for the separator, the logic is the same for both.  I'm not a fan of interspersing the separator because it doubles the intermediate memory usage, which could be a problem in reasonable use-cases.  (For example, joining an large array of integers by commas -- the stack manipulation would likely be the most expensive part.)

Another approach would be to just have an implementation that supports separators and have both the meta-op and `table.concat` use that internally, but that adds some complexity to the normal concat case.

src/thread/thread.rs
```diff
@@ -805,15 +805,60 @@ impl<'gc, 'a> LuaFrame<'gc, 'a> {
...
+        let top = self.state.stack.len();
+        debug_assert_eq!(top, *base + *stack_size);
+
+        self.fuel.consume(Self::FUEL_PER_CALL);
+
+        *expected_return = Some(LuaReturn::Meta(meta_ret));
+
+        // This does not need to consume fuel for each argument, as
+        // the arguments are used in-place on the stack and are not
+        // shifted.
```
Aeledfyr, line 857: Is this correct / reasonable?  Is the per-argument cost solely tracking the stack manipulation, or is it assumed to exist on every function call?

src/meta_ops.rs
Aeledfyr: I've changed it to do this in the latest version, though you should check over the method (`call_meta_function_in_place`, name subject to change) to make sure that it lines up with what you expect.

src/meta_ops.rs
Aeledfyr: Fixed in the latest version.

src/stdlib/table.rs
Aeledfyr: I'll add overflow checks on this, but it won't actually do much -- even if it doesn't overflow, it'll crash on allocation if it gets close.   Properly handing OOM conditions (and preventing them from Lua) will take a significant amount of work, so yeah, it's probably not something to worry about right now.

src/stdlib/table.rs
Aeledfyr: Oh, another minor difference is that for the concat metaop, the first half isn't a Sequence, but it has to be in the `table.concat` case, to be able to tail-call `table.unpack` and have it return into the concat implementation.  As such, some of the inner implementation can be reused (and some currently is), but each case has different control-flow requirements which make merging them difficult.

Aeledfyr:
What would be the current best approach to fuel consumption for concat?  Consume fuel proportional to the output size at each step?  (This runs into the issue of not being able to efficiently break up large allocations, though -- would need to repeatedly consume fuel / yield while doing nothing, and then once enough fuel has been consumed, make the full large allocation.)

src/stdlib/table.rs
Aeledfyr: Overflow checks have been added, and the estimation code is now shared between both versions.

src/meta_ops.rs
```diff
@@ -735,7 +739,105 @@ pub fn less_equal<'gc>(
...
+            Value::Number(_n) => 10,
+            Value::String(s) => s.as_bytes().len(),
+            _ => return Ok(None),
+        };
+        len = len
+            .checked_add(value_len)
+            .ok_or(MetaOperatorError::ConcatOverflow)?;
+    }
+    Ok(Some(len))
+}
```
kyren, line 775: This is so hyper specific that I would choose any other option than have this as a public method.

I would rather just repeat the code, or even just not estimate length at all.

Aeledfyr: Ah, whoops -- I didn't mean to make that fully public.  Would it be reasonable to make it `pub(crate)` instead, or should I just put it back into the two concat impls?  It's the simplest part to share between the two versions.

For now I'll make it `pub(crate)`, but I can inline it again if needed.  Though it may not be obvious that changes to this would also need to be reflected in `table.rs`... would it be better to move all of the concatenation code into `meta_ops.rs`, even the stuff that handles separators?

Aeledfyr: `estimate_concatenated_len` is no longer public, and I ended up moving the separated version into `meta_ops`, next to the normal version.  (It may not be the best place for it, since there's no meta-operator that uses it, but it's better to have the two variants of concat in the same place.)

src/thread/thread.rs
kyren: This is correct, it's solely tracking stack manipulation.

kyren:
I'm sorry it took me so long to merge this, I've had a rough week so I've been a little out of it.

I ended up moving the support code for async sequences used by meta ops into meta_ops.rs and making it private. I know this is the opposite of what we discussed before, but I didn't realize it was something ONLY used by meta_ops.rs, so this made more sense to me. This was the only meaningful feedback I had after sitting on it for a week (and it was a tiny change) so I just went ahead and merged the PR rather than bother you about it.

*Thank you* for all of your hard work. With this, every metamethod is implemented other than `__gc`!

Aeledfyr:
Thank you!  There's no problem with the wait on the merge (after all, I'm responding with a significantly longer delay; life happens).

On the async support code: I agree, but I think it would still be worth making something like it available in the future (potentially in `piccolo_util`), since it helps a lot with the verbosity of async sequences.  It would need more planning to determine which helpers/API would work best, and at this point I don't have enough usage examples to come up with a good design.

PR #89, opened 2024-07-03
Implement comparison metamethods, support them in `math.min`/`max`

Aeledfyr:
Implements the `__lt` and `__le` metamethods, and implements versions of `math.min` and `math.max` that support those metamethods.

This also changes the default comparison behavior to not automatically convert strings to numbers in comparisons, to match PRLua 5.4.

Along with #88, this implements the last of the missing metamethods.


kyren:
This looks really good, thank you!

PR #93, opened 2024-09-12
feat: basic bench harness

jprochazk:
Adds a basic Criterion-based benchmark harness which loads scripts from the `benches/scripts` directory and times their execution as if they were regular Rust functions.
The Criterion settings were tuned for fast runs, at the expense of greater run-to-run variability. It is highly recommended that you pin the benchmark to a specific CPU core to reduce the noise.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,7 @@
 [workspace]
 resolver = "2"
-members = [
-    "util",
-]
-default-members = [
-    ".",
-    "util",
-]
+members = ["util"]
+default-members = [".", "util"]
 
 [workspace.package]
 version = "0.3.3"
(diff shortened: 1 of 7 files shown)


Cargo.toml
Aeledfyr, line 1: Looks like this commit re-added the unintended formatting changes.

benches/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+This directory contains benchmarks for `piccolo`.
+
+## Running benchmarks
+
+```
+cargo bench
+```
+
+Note that it may take a while to compile, because `lto` is set to `fat`.
```
Aeledfyr, line 11: Since LTO has been disabled, this should probably be updated.

Cargo.toml
jprochazk: True! It's easy to miss, would be nice to have some kind of formatter setup for `.toml` files in this repository

benches/README.md
jprochazk: Yep, removed this line. On my machine a compile on a `cargo clean`-ed repository goes from 40s to 20s without `lto=fat`.

Cargo.toml
Aeledfyr: I can't mark these as comments as resolved, but feel free to resolve both (assuming Github lets you).

PR #94, opened 2024-09-14
Avoid monomorphizing a copy of the compiler for each `T: Read` in `Closure::load`

Aeledfyr:
The parser also has the same issue, but it isn't easily fixable without tradeoffs.  (Currently there is a minimum of 2 instantiations of the entire parser, once from the Lua stdlib and once for the user program, assuming it only passes byte slices to Closure::load.)

A temporary fix for the parser would be to have `Closure::load` read everything into a byte slice before invoking the parser, but a potential long-term solution would be to make the parser non-generic, so that users can use the parser directly without monomorphizing it at each usage site.


Aeledfyr:
Superseded by #117.

PR #95, opened 2024-09-14
Fix panic in concat with negative integers

Aeledfyr:
Currently string concatenation panics when used with integers ≤ 0 due to a bug in my use of ilog10.  This PR fixes that bug and adds a test for it.

Example case that triggers the bug:
```lua
local i = 0
local test = "Hello, World!" .. i
```


Aeledfyr:
Fixed in #106

PR #97, opened 2024-11-04
Implement `string.byte` and `string.char`

ginnyTheCat:
Add `string.byte` and `string.char` to the standard library.


ginnyTheCat:
As far as it seems the failing CI run is caused this change https://github.com/rust-lang/rust/commit/0747f2898e83df7e601189c0f31762e84328becb

PR #99, opened 2024-11-17
Check the size of piccolo's Thread, not std's.

ahmedcharles:


ginnyTheCat:
The only reason they fail under miri is that the sizes changed in the newest nightly version with which miri is being run. The only appropriate fix would be to adjust or remove the test all together.

ahmedcharles:
Do you know why the size changed with the nightly? Do you have a link to that RFC?

ginnyTheCat:
Jup. I noticed it when the CI build for my PR failed as well and it's caused by this PR https://github.com/rust-lang/rust/pull/123550.

What I don't quite understand is what the reason is piccolo checks whether these sizes are what they are, as from the little code I read piccolo doesn't seem to use `std::thread::Thread` anywhere else. What I'm assuming is that it happened to be a miss-import and actually the size of `piccolo::Thread` should've been tested instead (which is a pointer wide as well).

ahmedcharles:
Yes, it's supposed to check `piccolo::Thread`. I assume this was due to an IDE importing the wrong `Thread`.

kyren:
Thank you for the PR!

Yes, that is *definitely* supposed to check the size of piccolo's `Thread`, it's supposed to check to make sure it's a thin pointer.

PR #100, opened 2024-11-21
Change Debug for Table to be shallow.

### LINKED ISSUE #90: RawTable fmt::Debug implementation can overflow stack
When printing a table with `{:?}`, if it has __index set to itself, it overflows the stack by recursively printing itself

Not entirely sure what the desired behavior should be, especially since the cycles probably won't always be that simple? But right now, debug printing the wrong thing will cause it to crash 😬

ahmedcharles:
Fixes #90

### DIFF
--- a/src/table/raw.rs
+++ b/src/table/raw.rs
@@ -41,11 +41,16 @@ impl<'gc> fmt::Debug for RawTable<'gc> {
                 self.array
                     .iter()
                     .enumerate()
-                    .map(|(i, v)| (Value::Integer((i + 1).try_into().unwrap()), *v))
+                    .map(|(i, v)| {
+                        (
+                            Value::Integer((i + 1).try_into().unwrap()).debug_shallow(),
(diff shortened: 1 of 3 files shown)


Aeledfyr (changes requested):
Sorry about the long delay on reviewing; if you're still interested in working on this, there are some review comments below, but if not I can pick this up and finish it.

This looks reasonable, though it doesn't handle the case where the cycle is in the metatable:
```lua
a = {}
setmetatable(a, a)
```
The fix for that should just be a custom implementation of `Debug` for `TableState` (in `src/table/table.rs`).  It may be nice to have it show the first level of the metatable in that case?  (by directly printing the inner table, but printing the metatable's metatable as shallow?)

ahmedcharles:
>  but if not I can pick this up and finish it.

I'm ok with that.

Aeledfyr:
I'll do that then.  Thank you for the PR!

PR #103, opened 2025-02-06
Trivial typo fix

mkeeter:


mkeeter:
It looks like the CI failure is unrelated to this change:
```
error: package `home v0.5.11` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.77.2
```

PR #112, opened 2025-05-08
`Feature`: implemented (almost) all `io.*` and `file:*` functions in stdlib

reloginn:

# Features
- Implemented functions `io.close`, `io.flush`, `io.input`, `io.lines`, `io.open`, `io.output`, `io.read`, `io.tmpfile`, `io.type`, `io.write`, `file:flush`, `file:lines`, `file:read`, `file:seek`, `file:write`.
- Added tests in `io.lua` for all new functions.
- In `Cargo.toml`, the `tempfile` and `either` crates have been added. `tempfile` is used for `io.tmpfile`, and `either` is used as a helper in the `IoFile` type.


notxvilka:
@Aeledfyr since you are reviewing all PRs, could you please review this one as well?

Aeledfyr:
I can review this, but I'm only merging relatively small changes.  Larger changes like this will have to wait until kyren is available to review as well, which will likely be a while from now.

I'll do a first-pass review of the larger PRs (and see if there are sub-parts that can be split out to merge sooner), but it may be a while before they can be fully merged.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+                Value::String(filename) => {
+                    let io: Table = ctx.globals().get(ctx, "io")?;
+                    let open: Callback = io.get(ctx, "open")?;
+
+                    stack.replace(ctx, (filename, "r"));
+                    Ok(CallbackReturn::Call {
+                        function: open.into(),
+                        then: Some(BoxSequence::new(&ctx, InputOpenThen)),
+                    })
+                }
```
Aeledfyr, line 189: This can be simplified significantly by making the implementation of `open` a normal Rust closure---then, rather than having to call it through the globals table and callbacks/sequences, you can call it as a normal function.

```rust
let open = move |filename: &String, mode: &String| -> Result<IoFile, Error> {
    // ...
};
io.set_field(ctx, "open", Callback::from_fn(&ctx, |ctx, _, mut stack| {
    let (filename, mode) = stack.consume::<(String, Option<String>)>(ctx)?;
    let result = open(&filename, &mode);
    match result {
        // ...
    }
});
io.set_field(ctx, "input", Callback::from_fn(&ctx, |ctx, _, mut stack| {
    let file = stack.consume::<Option<Value>>(ctx)?;
    match file {
        Value::String(name) => {
            let file = open(&name, "r");
            // ...
         },
        // ...
    }
});
```
Note that you might have to clone the `open` closure if it captures state (like the io metatable); `let open_ref = open.clone()` should work for that.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+                Value::UserData(file) => {
+                    if let Ok(file) = file.downcast_static::<IoFile>() {
+                        IO_STATE.with(|state| state.replace_input(file.clone()));
+
+                        stack.replace(ctx, UserData::new_static(&ctx, file.clone()));
+                        Ok(CallbackReturn::Return)
```
Aeledfyr, line 195: This could return the original UserData, rather than creating a new userdata for it.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+                    if let Err(err) = output.close() {
+                        stack.replace(
+                            ctx,
+                            (Value::Nil, err.to_string(), err.raw_os_error().unwrap_or(0)),
+                        );
+                        return Ok(CallbackReturn::Return);
+                    }
+                    IO_STATE.with(|state| state.replace_output(IoFile::stdout()));
+                    stack.replace(ctx, true);
+                    return Ok(CallbackReturn::Return);
```
Aeledfyr, line 297: This could be removed, as it's the same as the code that will run after the if statement ends.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+    io.set_field(
+        ctx,
+        "read",
+        Callback::from_fn(&ctx, |ctx, _, mut stack| {
+            let input = IO_STATE.with(|state| state.input());
+
+            let file: Table = ctx.io_metatable().get(ctx, MetaMethod::Index)?;
+            let read: Callback = file.get(ctx, "read")?;
+
+            stack.into_front(ctx, Value::UserData(UserData::new_static(&ctx, input)));
```
Aeledfyr, line 355: Like `open`: `read`, `write` and `lines` should be implemented as a Rust function or closure that can be directly called here, rather than having to go through the global lookup/userdata/call setup.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+            for format in formats {
+                match file.read_with_format(ctx, format)? {
+                    Some(value) => {
+                        stack.into_back(ctx, value);
+                    }
+                    None => {
+                        stack.into_back(ctx, Value::Nil);
+                    }
+                }
+            }
```
Aeledfyr, line 670: Rather than collecting into an intermediate `Vec`, this should be done in one pass -- iterating through the format values on the stack, and replacing each one with the returned value.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+    file.set_field(
+        ctx,
+        "lines",
+        Callback::from_fn(&ctx, |ctx, _, mut stack| {
+            #[derive(Collect, Clone)]
+            #[collect(require_static)]
+            struct Lines {
+                position: RefCell<usize>,
+                file: IoFile,
+                formats: Vec<std::string::String>,
```
Aeledfyr, line 685: The formats should be parsed and stored as an enum, so that the rest of the implementation doesn't have to deal with possibly invalid format modes.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+                stack.consume::<(UserData, Option<String>, Option<i64>)>(ctx)?;
+            let file = if let Ok(file) = file.downcast_static::<IoFile>() {
+                file
+            } else {
+                return Err("bad argument #1 to 'seek' (file expected)"
+                    .into_value(ctx)
+                    .into());
+            };
+
+            let whence_str = whence.map(|s| s.to_str().unwrap_or("cur")).unwrap_or("cur");
```
Aeledfyr, line 796: This could be written as `whence.and_then(|s| s.to_str()).unwrap_or("cur")`, but it would be better to leave it was a byte string (rather than treating invalid utf8 as `"cur"`) -- so `whence.map(|s| s.as_bytes()).unwrap_or(b"cur")`.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+
+            match file.inner() {
+                Either::Left(left) => {
+                    let mut left = left.borrow_mut();
+                    if let Some(ref mut file) = *left {
+                        for value in values {
+                            file.write_all(value.as_bytes())?;
+                        }
+                    }
+                    stack.replace(ctx, UserData::new_static(&ctx, file.clone()));
```
Aeledfyr, line 896: This should return the existing file userdata, rather than creating a new one.

src/stdlib/io.rs
```diff
@@ -69,4 +88,831 @@ pub fn load_io<'gc>(ctx: Context<'gc>) {
...
+                } else {
+                    return Err("bad argument #1 to 'write' (file expected)"
+                        .into_value(ctx)
+                        .into());
+                }
+            } else {
+                return Err("bad argument #1 to 'write' (file expected)"
+                    .into_value(ctx)
+                    .into());
+            };
```
Aeledfyr, line 867: There should be a better way of managing the input arguments -- all of these errors have the exact same message, so it should be possible to shorten this?  (This also occurs in a number of other places.)

src/stdlib/io/file.rs
```diff
@@ -0,0 +1,323 @@
...
+                                    }
+                                    break;
+                                }
+                                _ => break,
+                            }
+                        }
+                    }
+                    Err(err) => return Err(err.to_string().into_value(ctx).into()),
+                }
+            }
```
Aeledfyr, line 108: Lua's readline only treats `\n` as newlines, so this can just read until it finds a `\n` byte.  I'd recommend using `BufRead::read_until` here instead, as it can handle it more efficiently than making a read for each byte.

src/stdlib/io/file.rs
```diff
@@ -0,0 +1,323 @@
...
+};
+
+use super::std_file_kind::StdFileKind;
+use either::Either;
+use gc_arena::Collect;
+
+fn read_from_any<'gc, W: Write + Read + Seek>(
+    ctx: Context<'gc>,
+    file: &mut W,
+    format: &str,
```
Aeledfyr, line 16: It would be good to make the format an enum, so that this function doesn't need to deal with parsing the format specifier.

src/stdlib/io/file.rs
```diff
@@ -0,0 +1,323 @@
...
+    ) -> io::Result<usize> {
+        file.seek(io::SeekFrom::Start(*position as u64))?;
+        match file.read(buf) {
+            Ok(n) => {
+                *position += n;
+                Ok(n)
+            }
+            Err(err) => Err(err),
+        }
+    }
```
Aeledfyr, line 32: The offset into the file shouldn't need to be manually tracked -- the files track those internally, and Rust already exposes the calls to change the offset through the Seek trait.

src/stdlib/io/file.rs
```diff
@@ -0,0 +1,323 @@
...
+                .ok()
+                .map(Value::from)
+                .or_else(|| s.parse::<f64>().ok().map(Value::from))
+            {
+                Some(value) => Ok(Some(value)),
+                None => {
+                    *position = start_position;
+                    Ok(None)
+                }
+            }
```
Aeledfyr, line 180: This should use the existing number parsing logic from the lexer (like `compiler::string_utils::{read_integer, read_float}`) so that it supports Lua's number syntax.

src/stdlib/io/file.rs
```diff
@@ -0,0 +1,323 @@
...
+                read_from_any(ctx, file, format, &mut read_state)
+            }
+            Either::Right(kind) => match kind {
+                StdFileKind::Stdin => {
+                    let mut read_bytes = self.read_position.borrow_mut();
+                    let mut stdin = io::stdin();
+                    let mut buf = Vec::new();
+                    stdin.read_exact(&mut buf)?;
+                    let mut cursor = Cursor::new(buf);
+                    read_from_any(ctx, &mut cursor, format, &mut read_bytes)
```
Aeledfyr, line 305: This approach won't work, since always reads a fixed amount of data (currently 0 bytes).  I assume this was a work-around for stdin/stdout not supporting Seek, but `read_from_any` just needs to be changed to support non-seekable files.

Aeledfyr (changes requested):
> All functions are based on and fully comply with the Lua 5.3 manual

Note that piccolo is generally targetting Lua 5.4 rather than 5.3, but there don't seem to be many differences in the io/file APIs.

I've gone through and done a first-pass review; there are a number of things that can be simplified, but the main things are to reuse the inner implementations directly, rather than going through Lua calls to use them (for the io functions that call file methods), and to try to simplify the input checking / error handling.

rib:
> Another high-level thing that needs to be considered is how to handle sandboxing this, and how to expose that to the user. The main things we'd need to achieve for reasonable sandboxing:
> 
> * Allowing the host to limit access to stdin/stdout/stderr, and/or replace them with custom files/streams
> * Allowing the host to limit what files can be opened, and what operations can be done on them
>
> One approach that might make this possible would be to have the user provide the system interface through a trait implementation -- something like this:

Yeah, this suggestion to add some kind of trait abstraction around I/O or OS details sounds good, to avoid assuming it's safe/desirable to expose the host OS / filesystem to Lua code.

I've been playing with using piccolo recently and at least for my use case it's a feature that piccolo doesn't support the io APIs which allow code to interact with the host platform.

Issue #120, opened 2025-06-16
Repl multiline functionality doesn't work (downcast_ref::<ParseError> is None, not Some(ParseError{..})).


Phraeyll:
Multiline functionality in the repl no longer works, for example, defining a function or the body of an "if" statement.  It looks like the changes to cleanup error handling in  ecb63e1 depend on the CompilerError variants to have the attributes set in that commit to specific strings:

```
#[derive(Debug, Error)]
pub enum CompilerError {
    #[error("parse error")]
    Parsing(#[from] compiler::ParseError),
    #[error("compile error")]
    Compilation(#[from] compiler::CompileError),
}
```

Commit  bc7d586 reverts the variants back to transparent:
```
#[derive(Debug, Error)]
pub enum CompilerError {
    #[error(transparent)]
    Parsing(#[from] compiler::ParseError),
    #[error(transparent)]
    Compilation(#[from] compiler::CompileError),
}
```

I'm not sure what the correct fix is  Either application code (like the repl example) needs to change to downcast the error to "CompilerError" rather than "ParseError", and pattern match with the more complicated object nesting, or the error variants in the library code need to be changed back to what they were in  ecb63e1.  Perhaps #[error(transparent)] was a mistake recommitting an older version?

Phraeyll:
repl error for reference:
> if (true) then
runtime error: parse error at line 1: unexpected end of token stream, expected End

Aeledfyr:
Thanks for the report; I've fixed it in #123.

(`#[repr(transparent)]` was re-added to fix a regression from the older error cleanup, where all parse/compile errors only printed as "compile error" or "parse error" with no addition information.  However, another part of #52 fixed that, so reverting repr(transparent) now only makes the error messages slightly worse.)

PR #121, opened 2025-07-08
Support capturing Error backtraces and `debug.traceback`

rib:
This adds optional backtraces to `Error::Runtime` and `Error::Lua` which are automatically captured when step() sees an error.

When '#' / alternative formatting is specified, the `Display` implementation for `Errors` will pretty-print any backtrace like:

```
Error: lua error: error: test error from rust
stack traceback:
  <test>:4 in <function 'error_callback' at line 1>
    arguments:
      err_msg: test error from rust
  <test>:9 in <function 'baz' at line 7>
    arguments:
      a: test
      b:  error from rust
  <test>:13 in <function 'bar' at line 12>
    arguments:
      x: test
      ...:  error from rust
  <test>:19 in <function 'foo' at line 16>
    arguments:
      arg1: test
  <test>:22 in <chunk>
```

There is also a `pretty_print()` method that can be given a source map (that maps chunk names to source code) that will include a line of source code for each frame, like:

```
Error: lua error: error: test error from rust
stack traceback:
  <test>:4 in <function 'error_callback' at line 1>: `error(err_msg)`
    arguments:
      err_msg: test error from rust
  <test>:9 in <function 'baz' at line 7>: `error_callback(a .. b)`
    arguments:
      a: test
      b:  error from rust
  <test>:13 in <function 'bar' at line 12>: `baz(x, ...)`
    arguments:
      x: test
      ...:  error from rust
  <test>:19 in <function 'foo' at line 16>: `bar(arg1, " error from rust")`
    arguments:
      arg1: test
  <test>:22 in <chunk>: `foo("test")`
```

The `debug.traceback` implementation doesn't try and conform to the same traceback format as PUC-Lua, or any other Lua implementation, but should honour the optional `thread`, `msg` or `level` arguments.


rib:
It looks like the Circle CI / Miri failures also show up on master and aren't introduced by this PR

src/stdlib/debug.rs
```diff
@@ -0,0 +1,88 @@
...
+                }
+            } else {
+                // If no thread is specified, default to the current running thread.
+                crate::thread::backtrace(
+                    exec.upper_frames(),
+                    stack.borrow_thread_view(),
+                    &[],
+                    Some(level0),
+                )
+            };
```
Aeledfyr, line 57: The current thread and fallback cases here seem like they could be merged, though I'm not sure how to do it cleanly.  Here's one approach:
```rust
if let Some(t) = thread.filter(|t| t != exec.current_thread().thread) {
    // other thread
} else {
    // cur thread
}
```

<details>
<summary>
(It's simpler with let chains, but I don't know kyren's stance on bumping the MSRV.)
</summary>

```rust
if let Some(t) = thread && t != exec.current_thread().thread {
    // other thread
} else {
    // cur thread
}
```
</details>

tests/backtrace.rs
```diff
@@ -0,0 +1,187 @@
...
+    source_map.insert("test_backtrace.lua".to_string(), SCRIPT.to_string());
+
+    if let Err(
+        ref err @ ExternError::Lua {
+            ref error,
+            ref backtrace,
+            ..
+        },
+    ) = result
+    {
```
Aeledfyr, line 30: These might be easier to structure as
```rust
let Err(...) = result else {
    panic!("message");
};
```
mainly to keep the panic closer to the source

Aeledfyr (commented):
I haven't had a chance to look through this fully, but it looks really good!  Thank you!

The only design question here (meaning I'll need to wait for approval from kyren for merging) is whether to automatically attach backtraces to errors, and how that influences Lua compatibility.  There are a few discussions on the piccolo discord channel on it, [such as this one](https://discord.com/channels/865004050357682246/1171939909272092712/1174087855409201284), but the general gist is that making it compatible with base Lua is hard because Lua generally converts errors to strings and makes passing through metadata difficult.

(I'll also see if I can fix the CI issue, should be doable on my end.)

src/stdlib/debug.rs
rib: yeah I think when I first wrote this without running `cargo fmt` the `backtrace()` call was on one line and it seemed more acceptable to just have the duplicate paths for the current thread case, but yeah it'd be nicer to avoid that.

src/error.rs
```diff
@@ -298,15 +358,36 @@ impl StdError for ExternError {
...
+            ExternError::Lua { error, .. } => error,
             ExternError::Runtime(err) => err.root_cause(),
         }
     }
+
+    pub fn pretty_print(
+        &self,
+        writer: &mut impl fmt::Write,
+        source_map: Option<&SourceMap>,
+    ) -> Result<(), fmt::Error> {
```
Aeledfyr, line 370: This may be easier to use if it returns `impl fmt::Display` rather than taking a Write implementation; for an example see https://github.com/kyren/piccolo/blob/ce709eb1dae5c543cbc78e7e12bb80249d88c55f/src/value.rs#L55-L59
(Actually, `pretty_print_error_with_backtrace` could itself return `impl fmt::Display` to reduce code duplication from monomorphization.)

Aeledfyr (commented):
With this PR, RuntimeError and LuaError store the backtrace internally, but ExternLuaError doesn't, and puts it in the ExternError struct.  It seems like these would be easier to manage if they were all the same?  In addition, that would allow preserving the conversion from `LuaError` to `ExternLuaError`.

**Edit:** Also, `StashedError` doesn't preserve the backtrace, which may make errors lossy in async_sequence-based callbacks.

Also, the current approach in piccolo for error formatting is using `{:#}` on RuntimeError in the Display impl for Error/ExternError, which the RuntimeError passes to `anyhow` to generate a single-line error chain for the errors.  ("runtime error: operator error: could not add values of type table and number").

This PR currently replaces that with manual chain printing (`#[error("operator error: {0}")]`) in VMError, since the alternate mode was reinterpreted to mean printing the backtrace, but it may be better to fix that by always enabling alt mode when printing the anyhow error (in `impl Display for RuntimeError`, and in `pretty_print_error_with_backtrace`).

(I also don't know the best way of exposing the enabling of backtrace printing to the user -- there are four general options for formatting (`{}`, `{:#}`, `{:?}` and `{:#?}`), and anyhow treats all of them differently, but I'm not sure what the best approach for `piccolo` would be.)

Aeledfyr:
I've fixed the CI issue and an error downcasting bug (#120, fixed in #123) on master; I don't think there should be any conflicts, but it's worth checking.  It's also worth checking if the error changes from this PR would re-introduce #120, since the error downcasting is tied pretty closely to the error format.

src/thread/thread.rs
```diff
@@ -529,6 +540,100 @@ impl<'gc> ThreadState<'gc> {
...
+    thread_stack: &[Thread<'gc>],
+    top_frame: Option<BacktraceFrame>,
+) -> Vec<BacktraceFrame> {
+    let mut trace_frames = Vec::new();
+
+    if let Some(frame) = top_frame {
+        trace_frames.push(frame);
+    }
+    for i in (0..frames.len()).rev() {
+        if let Some(frame) = frames.get(i) {
```
Aeledfyr, line 555: This could be simplified to
```rust
for frame in frames.iter().rev() {
    // ...
}
```
(`i` is unused elsewhere, but if you needed it, it would be `for (i, frame) in frames.iter().enumerate().rev() { ...`)

src/stdlib/base.rs
```diff
@@ -103,9 +104,9 @@ pub fn load_base<'gc>(ctx: Context<'gc>) {
             if stack.get(0).to_bool() {
                 Ok(CallbackReturn::Return)
             } else if stack.get(1).is_nil() {
-                Err("assertion failed!".into_value(ctx).into())
+                Err(LuaError::from("assertion failed!".into_value(ctx)).into())
             } else {
-                Err(stack.get(1).into())
+                Err(LuaError::from(stack.get(1)).into())
             }
```
Aeledfyr, line 110: `LuaError::from` is unnecessary here and in the definition of `error`

Aeledfyr:
Okay, I've checked with kyren;
- Auto-capturing backtraces is fine, and should be the default; we don't really need config for the initial version
- The `Vec<BacktraceFrame>` should have a wrapper struct; then `pretty_print_error_with_backtrace` would become a method on that struct
- It might be slightly cleaner to store the backtraces in the `Error` rather than in each variant, but that means turning it into `struct Error { kind: ErrorKind, backtrace: Option<Vec<BacktraceFrame>> }` for each of `Error`, `ExternError`, and `StashedError`.  I don't know if this would actually be simpler, or worth the effort.

In the longer term, an executor will likely have an error hook function called whenever an error is initially thrown (when a new Frame::Error is pushed when not currently handling an error); then that hook could capture the backtrace and attach it to the error without having to have a backtrace field in LuaError/RuntimeError, by using a custom rust error type.  This would allow the user to control how the backtraces are captured, as well as if/how they're exposed to Lua.

PR #125, opened 2025-07-11
Update to rand 0.9

Jengamon:
Bump code to use rand 0.9 by translating API calls:
- `gen*` -> `random*`
- `from_entropy` -> `from_os_rng`


Aeledfyr:
We may want to make `rand` optional, or directly expose RNG seeding through the API before upgrading to rand 0.9; that bumps getrandom to 0.3, which is more difficult to get working on WASM.  (`getrandom` is an annoyingly painful dependency to manage.)

It should be straightforward to split `load_random` (and `load_math`) into variants that initialize from a seed and variants that use OS apis, and put the ones that require `getrandom` behind a feature flag.  I may go and fix that at some point when I have time, but until then I'm personally stuck on rand 0.8.

Jengamon:
Would making the rng an argument be a path we'd want to go down?

Different applications might want different random number properties, and then we could turn off all default features, and have a feature that provides the old `load_math` interface with os-seeded rng.

Jengamon:
> We may want to make `rand` optional, or directly expose RNG seeding through the API before upgrading to rand 0.9; that bumps getrandom to 0.3, which is more difficult to get working on WASM. (`getrandom` is an annoyingly painful dependency to manage.)
> 
> It should be straightforward to split `load_random` (and `load_math`) into variants that initialize from a seed and variants that use OS apis, and put the ones that require `getrandom` behind a feature flag. I may go and fix that at some point when I have time, but until then I'm personally stuck on rand 0.8.

I haven't been having too many problems after sticking
```toml
[target.wasm32-unknown-unknown]
rustflags = '--cfg getrandom_backend="wasm_js"'
```
into `.cargo/config.toml` relative to the crate I want to compile for wasm (along with the feature addition in Cargo.toml)

**EDIT**: they tell you this on the crate docs page ignore me

PR #133, opened 2025-10-03
Do not call `__index` or `__newindex` from `table.insert`

csnover:
PUC-Lua does not do this and real world scripts rely on this not happening as they call `table.insert` from `__newindex`.


csnover:
I was misled by a Lua playground I used to try to verify the PUC-Lua behaviour which I thought was running 5.4 but actually was running 5.2. :-( Since I am trying to support scripts that were written for an earlier Lua version, this change only makes sense if piccolo gets some compatibility flag, and I don’t know if that is in scope for what you are trying to do. Sorry about the noise!

PR #134, opened 2025-10-03
Support string method calls

csnover:
Technically, to be 100% compatible, this should be calling through a metatable assigned to all strings which has __index pointing to the string global table. This intermediate approach works for all sane use cases without requiring so much work. (It is not possible to *set* a metatable on a string in PUC-Lua, but it is at least possible to call `getmetatable` on a string and then mess with it.)


ProgramCrafter:
I implemented the 100% compatible approach at https://github.com/ProgramCrafter/piccolo/commit/382e4ccd58b1492e3ec59a3808717e2cf0074aa2

PR #135, opened 2025-10-03
Show more detailed compiler error messages

csnover:
On the last commit to master, transparent compiler errors were reverted, which, as noted, made the error messages a little worse. But what if it is possible to have the same good error messages too? :-)


csnover:
Whoops! I didn’t notice that this was only working correctly in conjunction with PR #121, which I have been running locally. I thought it was weird that the error messages were not being duplicated with this one weird fix :-) Well, if that PR ever lands, then this change should also be beneficial.

PR #138, opened 2025-12-18
Avoid pessimising integer arithmetic with string operands

csnover:
This stops string-integer operands causing `Value` to convert to floating point unnecessarily.


csnover:
The CI failure is unrelated to this patch, a dependency increased its MSRV so now the build is failing.

PR #148, opened 2026-09-20
Upgraded Gc Arena to latest Git (0.7+) and edition 2024

Phraeyll:
I managed to upgrade this to the latest GC arena.  Due to changes in the library, there are issues in the following areas, even though all tests pass.  I'm not too familiar with git or Github, so I put the changes inside a different branch.  I'm not sure if that's an issue or not.

- **MetricsAlloc is gone** - Should this be implemented in piccolo instead? Is it needed, or too complex to reason about as it was in gc-arena?  Many functions no longer have need/use for the ctx parameter now, and have been removed.

- **"total_gc_bytes" just replaced with "total_gc_count"** - This almost certainly means logic errors (especially in the the lua stdlib "collectgarbage", but also in the public rust API of the Lua struct "total_memory")

```rust
    ctx.set_global(
        "collectgarbage",
        Callback::from_fn(&ctx, move |ctx, _, mut stack| {
            match stack.consume::<Option<String>>(ctx)? {
                Some(arg) if arg == "count" => {
                    stack.into_back(ctx, ctx.metrics().total_gc_count() as f64 / 1024.0);
                }
                Some(_) => {
                    return Err("bad argument to 'collectgarbage'".into_value(ctx).into());
                }
                None => {}
            }
            Ok(CallbackReturn::Return)
        }),
    );
```

- **The dyn string interner in string.rs no longer removes elements on comparison failure** - I'm not really sure why it did before, but it cannot do it now as hashbrown has been upgraded and no longer have access to RawTable and erase, and also typical rust iterator lifetime rules.  Maybe this can be implemented via hashbrown's "extract_if" but I couldn't figure it out.

before:
```rust
    fn intern(self, mc: &Mutation<'gc>, s: &[u8]) -> String<'gc> {
        // SAFETY: If a new string is added, we call the write barrier.
        let mut dyn_strings = unsafe { self.0 .0.unlock_unchecked() }.borrow_mut();

        // SAFETY: The RawTable outlives the iterator
        unsafe {
            for bucket in dyn_strings.iter_hash(str_hash(s)) {
                let (key, _) = *bucket.as_ref();
                if let Some(st) = key.upgrade(mc).map(String::from_inner) {
                    if st == s {
                        return st;
                    }
                } else {
                    // SAFETY: it is okay to erase items yielded by the iterator.
                    dyn_strings.erase(bucket);
                }
            }
        }

        // SAFETY: We are going to modify the dyn_strings table, so call the write barrier.
        Gc::write(mc, self.0);

        let s = String::from_slice(mc, s);
        dyn_strings.insert(
            s.stored_hash(),
            (Gc::downgrade(s.into_inner()), s.stored_hash()),
            |(_, hash)| *hash,
        );

        s
    }
```
after:
```rust
    fn intern(self, mc: &Mutation<'gc>, s: &[u8]) -> String<'gc> {
        // SAFETY: If a new string is added, we call the write barrier.
        let mut dyn_strings = unsafe { self.0.0.unlock_unchecked() }.borrow_mut();

        for (k, _) in dyn_strings.iter_hash(str_hash(s)) {
            if let Some(st) = k.upgrade(mc).map(String::from_inner) {
                if st == s {
                    return st;
                }
            }
        }

        // SAFETY: We are going to modify the dyn_strings table, so call the write barrier.
        Gc::write(mc, self.0);

        let st = String::from_slice(mc, s);
        dyn_strings.insert_unique(
            st.stored_hash(),
            (Gc::downgrade(st.into_inner()), st.stored_hash()),
            |(_, hash)| *hash,
        );

        st
    }
```

At least there are less unsafe blocks due to using HashTable instead of HashMap and RawTable, which was noted as a benefit in the comments.  This was required due to upgrade to hashbrown 0.15+ (0.17).  I didn't really notice anything else during this upgrade.  This mostly meant adding <'gc> to the collect trait, using dyn_collec! for Sequence trait to make it object safe, and switching to use hashbrown's HashTable.

If this is too hard to reason about, in retrospect, I think it might be better to look at smaller patches like:
1. Remove/Replace MetricsAlloc
2. Upgrade hashbrown to 0.17 and use HashTable
3. Finally upgrade GC-Arena (lifetime proliferation, and dyn_collect!)


Phraeyll:
Oh, and as for upgrading the edition to 2024, I upgraded rand to 0.10 (which includes the SmallRand feature by default) and used the renamed function (gen -> random) as the gen is a keyword reserved in 2024. I also needed to add explicit unsafe block to unsafe functions which internally used unsafe functions, mainly in freeze.rs.

